PR #11, opened 2023-11-02
Improve memory management and fix #8

### LINKED ISSUE #8: Unexpected exit when enable stage 1 translation in el2
When enable stage 1 translation in el2, create a gic-demo will get an error. Now I have annotated two lines in main.rs, fix it later.
image

KouweiLee:
There are some works I have done:
1. Fix issue #8 so that it works well when enable stage 1 translation in el2. 
2. Modify the address mapping in el2 and root cell, remove the root cell's memory area in el2 according to jailhouse.
3. Add the config address mapping to el2 when create a cell.
4. Fix all the warnings when compile project.
5. Use `cargo fmt` to improve code style


.vscode/settings.json
KouweiLee, line 1: These lines are from rcore, they can prevent error reporting in vscode

src/arch/aarch64/s1pt.rs
KouweiLee, line 1: Why did the entire file change? How can I avoid it?

src/arch/aarch64/s2pt.rs
KouweiLee, line 1: Why did the entire file change? How can I avoid it?

src/hypercall/mod.rs
KouweiLee, line 1: After enable stage 1 translation in el2, the config address is ipa which needed to change to pa and map to el2 address space. So according to jailhouse, I map this address to a temporary region.

src/cell.rs
```diff
@@ -201,13 +196,13 @@ impl Cell {
     }
 
     pub fn config(&self) -> CellConfig {
```
KouweiLee, line 198: `fn config` should check whether stage 1 translation in el2 is enabled. Otherwise it will get an fault address which leads to issue #8

Inquisitor-201 (approved):

PR #13, opened 2023-11-13
MMIO Handlers

Inquisitor-201:
1. Added MMIO handlers for GICD and GICR
2. Implemented security protection to prohibit cross-cell IRQ operations
3. Added subpages MMIO handlers
------------------------------------------------------------------------------------------------
Now we can boot a non-root Linux based on another serial port and another disk file system.


KouweiLee (approved):
Well done!!! Good job!!!

PR #14, opened 2023-11-15
Finish destroy a cell

KouweiLee:
Now we can destroy a cell including gic-demo and linux. But there are some mmio unhandled errors after destroying cell on my qemu. It should be tested to check whether it is the fault of qemu.

### DIFF
--- a/src/cell.rs
+++ b/src/cell.rs
@@ -5,6 +5,7 @@ use spin::RwLock;
 use crate::arch::Stage2PageTable;
 use crate::config::{CellConfig, HvCellDesc, HvConsole, HvSystemConfig};
 use crate::control::{resume_cpu, suspend_cpu};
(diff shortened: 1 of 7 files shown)


src/cell.rs
```diff
@@ -234,17 +236,19 @@ impl Cell {
...
     pub fn gpm_activate(&self) {
         unsafe { self.gpm.activate() }
     }
-
+    /// Query an ipa from cell's stage 2 page table to get pa.
     pub fn gpm_query(&self, gpa: GuestPhysAddr) -> usize {
         unsafe { self.gpm.page_table_query(gpa).unwrap().0 }
     }
-
+    /// Map a mem region to a cell. \
```
KouweiLee, line 249: I add some doc annotates to some public functions. I'm confused when there aren't any comments about the function.

Inquisitor-201:
Awesome!

Inquisitor-201 (approved):

PR #20, opened 2023-12-10
rename rvm to hvisor

KouweiLee:
1. rename rvm to hvisor
2. add a new patch: hvisor.patch, which replaces jailhouse.patch
3. remove demo, which is useless.
4. change `aarch64-cpu` to official version


Cargo.toml
```diff
@@ -13,7 +13,7 @@ bit_field = "0.10"
 numeric-enum-macro = "0.2"
 buddy_system_allocator = "0.8"
 tock-registers = "0.8"
-aarch64-cpu = {git = "https://github.com/likey99/aarch64-cpu"}
+aarch64-cpu = "9.4.0"
```
KouweiLee, line 16: The newest `aarch64-cpu` has added what likey wants to add. So just use the official version.

Inquisitor-201 (approved):

PR #78, opened 2024-10-30
add aia=aplic-imsic

SanchezDorso:
Add AIA and solve the TAB Completion Problem


Inquisitor-201:
Please pull the newest 'dev' branch and resolve conflicts in these files:
Cargo.toml
Makefile
src/device/irqchip/mod.rs
src/platform/mod.rs

src/arch/riscv64/zone.rs
```diff
@@ -70,4 +107,6 @@ impl Zone {
 pub struct HvArchZoneConfig {
     pub plic_base: usize,
     pub plic_size: usize,
+    pub aplic_base: usize,
```
KouweiLee, line 110: hvisor-tool's riscv HvArchZoneConfig is not equivalent to here.

KouweiLee (approved):
Please rename your pr title, and handle the problem I mentioned.

rust-toolchain.toml
```diff
@@ -1,5 +1,5 @@
 [toolchain]
 profile = "minimal"
 # use the nightly version of the last stable toolchain, see <https://forge.rust-lang.org/>
-channel = "nightly-2023-12-28"
+channel = "nightly-2024-04-28"
```
enkerewpo, line 4: is this version upgrade a must-do? Because of codegen bug in the rust LoongArch target, using a recent nightly version of rust will cause hvisor on LoongArch failed to initialize on all cpus

enkerewpo: Because rust nightly frequently changes the code generation and api conventions, there are risks that upgrading the nightly version will introduce new problems on aarch64, riscv64 and loongarch64.

SanchezDorso: Not necessary.I set nightly-2024-04-28 just because when run CI workflow ,it failed due to :

error: failed to compile `cargo-binutils v0.3.6`, intermediate artifacts can be found at `/tmp/cargo-installitqA76`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.

Caused by:
  package `cargo-platform v0.1.9` cannot be built because it requires rustc 1.78 or newer, while the currently active rustc version is 1.77.0-nightly
  Try re-running cargo install with `--locked`

enkerewpo: ok, then I think we need to rewind the version to 2023-12-28

src/arch/riscv64/zone.rs
SanchezDorso: hvisor-tool has been modified  to be equivalent to here.

liulog:
I can also run in qemu-system-riscv64 with aia.

image

PR #91, opened 2025-01-01
Add zcu102 platform support.

ForeverYolo:


scripts/zcu102-aarch64-fit.its
enkerewpo, line 1: this `root_dtb`'s addr should be sync with the platform definition's `ROOT_ZONE_DTB_ADDR`

src/platform/zcu102_aarch64.rs
enkerewpo, line 1: this `ROOT_ZONE_DTB_ADDR` is not equal to FIT's its source code. Could you fix this, ty

enkerewpo (approved):

PR #101, opened 2025-02-11
aarch64: fix root linux not booting

enkerewpo:
## AARCH64
fix: set aarch64 makefile QEMU memory size to 3G to cover `linux1.dts`'s memory region, this will fix the problem of root linux not booting

## LOONGARCH64
fix: remove loongarch cfg control in hypercall enum to avoid compile error

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,8 +6,8 @@ PORT ?= 2333
 MODE ?= debug
 OBJCOPY ?= rust-objcopy --binary-architecture=$(ARCH)
 KDIR ?= ../../linux
-FEATURES ?= platform_zcu102,gicv2
-BOARD ?= zcu102
+FEATURES ?= platform_qemu,gicv3
+BOARD ?= qemu
 
 ifeq ($(ARCH),aarch64)
     RUSTC_TARGET := aarch64-unknown-none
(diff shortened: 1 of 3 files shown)


ZhongkaiXu:
I encountered the same issue when attempting to run hvisor. The vm traps into the mmio handler when trying to access PCI ECAM. Upon tracing the handler's behavior, I found that hvisor fails to return when attempting to use `mmio_perform_access`. I suspect that the page table entries related to ECAM in hvisor have been modified. Due to the high address range of ECAM, it might have been overlooked by the modifier. We're currently verifying this.

KouweiLee (approved):

KouweiLee:
> I encountered the same issue when attempting to run hvisor. The vm traps into the mmio handler when trying to access PCI ECAM. Upon tracing the handler's behavior, I found that hvisor fails to return when attempting to use `mmio_perform_access`. I suspect that the page table entries related to ECAM in hvisor have been modified. Due to the high address range of ECAM, it might have been overlooked by the modifier. We're currently verifying this.

@dallasxy will solve this problem by improving page table of hvisor in next pr, so this pr could merge and close.

PR #120, opened 2025-03-04
fix: temporarily disable the problematic subtests

CHonghaohao:
Comment out the tests related to "zone shutdown" in the tstart.sh and tresult.sh files


enkerewpo (approved):

enkerewpo:
LGTM, thank you. we need to deal with the zone shutdown issue in the future

PR #125, opened 2025-03-19
[Fix] fix qemu-aarch64 config & dts

Inquisitor-201:
1. Switch CPU to Cortex-A72 in platform and dts files.
2. Add gic fields in qemu-gicv3 configs.
3. Improve config_size error message in hypercall/mod.rs.


enkerewpo (approved):

enkerewpo:
LGTM

PR #126, opened 2025-03-20
[fix,license] force clean when build target board changed, added Mulan license header for project files

enkerewpo:
when ARCH or BOARD changed, force clean everything to avoid build bugs

### DIFF
--- a/platform/aarch64/qemu-gicv3/board.rs
+++ b/platform/aarch64/qemu-gicv3/board.rs
@@ -13,21 +13,6 @@
 //
 // Authors:
 //
-// Copyright (c) 2025 Syswonder
-// hvisor is licensed under Mulan PSL v2.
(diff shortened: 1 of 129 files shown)


enkerewpo:
updated license header, please view <https://gist.github.com/enkerewpo/e3c44cf5e6f857c39785a7408f7a9e23> for detail

platform/aarch64/qemu-gicv3/board.rs
KouweiLee, line 1: Why there are two times of header???

enkerewpo: weird, maybe some lagging when using git plugin in vscode through ssh, will be fixed immediately

enkerewpo:
some headers have been manually fixed

KouweiLee (approved):

PR #128, opened 2025-03-26
Update virtio-cfg for board qemu-gicv3

Inquisitor-201:


enkerewpo (approved):

src/arch/aarch64/zone.rs
```diff
@@ -50,7 +50,7 @@ impl Zone {
                     );
                 }
                 _ => {
-                    panic!("Unsupported memory type: {}", mem_region.mem_type)
```
KouweiLee, line 53: Why change here?

Inquisitor-201: 用户zone的配置错误不应该导致内核panic

KouweiLee: I will fix this issue soon.

Issue #129, opened 2025-03-26
riscv64: current vPLIC's logic has bugs


liulog:
Due to early-stage development and verification requirements, the simulation of the vPLIC was kept very simple, but this has led to some issues.

In the Linux system, there have been multiple instances where no output is generated during execution.

Upon investigation, I suspect  bugs in the **vPLIC's ​claim/complete mechanism** simulation. 

For example, when two interrupts arrive almost simultaneously and are claimed in succession, completing the first interrupt (i.e., clearing ​hvip.VSEIP) may result in an issue resembling **​interrupt loss**.

liulog:
I have added new vPLIC struct in my local code.
It works well. I will check more.
If I check done, I will submit a PR.

PR #130, opened 2025-04-01
riscv64: add vplic struct

liulog:
This PR is about #129.
- Add vPLIC's struct per zone. 
- vPLIC's emulation becomes more reasonable.
- Fix the bug mentioned before.

The AIA's code should be modifyied after.

Now qemu-plic's code works well.

image


liulog:
@SanchezDorso please check it.

SanchezDorso:
I can also run in qemu-system-riscv64 with plic.

SanchezDorso (commented):
Have checked it and I can also run in qemu-system-riscv64 with plic.
2025-04-02 23-03-19屏幕截图

KouweiLee (approved):

Issue #131, opened 2025-04-05
shutdown zonex can't shutdown zonex's first cpu


KouweiLee:
I find this rcu problem is related to one zonex's cpu being unable to shutdown. Probably the bug is about gic.

Issue #119, opened 2025-03-04
systemtest unstable because of rcu issue


enkerewpo:
The PR that added system tests for hvisor passed all CI checks, including the system tests themselves. However, after merging into the dev branch, the CI failed. The report indicates that the system test is unstable and may pass on a retry. This instability could be related to a previous RCU issue (noted by Honghao Cheng).

enkerewpo:
```
...
[INFO  0] (hvisor::memory::mm:97) region.start: 0x8000004000
[INFO  0] (hvisor::zone:210) zone cpu_set: 0b1100
[INFO  2] (hvisor::event:100) cpu 2 wakeup
[INFO  2] (hvisor::arch::aarch64::cpu:155) cpu 2 started
00:00:02 WARN  virtio_console.c:107: char device redirected to /dev/pts/0
Script started, file is /dev/null
# bash
root@(none):/home/arm64# dmesg | tail -n 3 | awk -F ']' '{print $2}' > ./test/tes
stresult/test_zone1_start.txt
root@(none):/home/arm64# [    5.515636] random: fast init done
./test/textract_dmesg.sh ./test/testresult/test_zone1_st
tart.txt
root@(none):/home/arm64# ./screen_linux2.sh
00:00:02 WARN  virtio_console.c:107: char device redirected to /dev/pts/0
/dev/pts/0

bash
# bash: cannot set terminal process group (-1): Inappropriate ioctl for device
bash: no job control in this shell
root@(none):/# cd /home/arm64
root@(none):/home/arm64# ls | grep zone1.txt
zone1.txt
root@(none):/home/arm64# 
[detached from 196.screen_linux2]
root@(none):/home/arm64# echo "zone1.txt" > ./test/testresult/test_zone1_ls.txt
root@(none):/home/arm64# ./hvisor zone list > ./test/testresult/test_zone_list2.t
txt
root@(none):/home/arm64# ./hvisor zone shutdown -id 1
[INFO  0] (hvisor::hypercall:240) handle hvc zone shutdown, id=1
[INFO  3] (hvisor::arch::aarch64::cpu:168) cpu 3 idle
[INFO  3] (hvisor::arch::aarch64::cpu:189) cpu 3 started from parking

(stuck forever)

```

CHonghaohao:
<img width="399" alt="Image" src="https://github.com/user-attachments/assets/c11641d8-1e78-41cb-a321-b47b41b5f1ed" />

The stuck information is as shown above

[...]

PR #132, opened 2025-04-05
add config check and add zone error handler to avoid panic

KouweiLee:
1. add config check to ensure compatiblity with hvisor-tool, so each time you change zone_config, please add CONFIG_MAGIC_VERSION by 1.
2. Improve shutdown zonex, now root zone's cpu doesn't need to wait for all cpus to be shutdown. Otherwise once a zonex's cpu is failed to shutdown, root zone will get an error `rcu`(like #119)
3. add zone error handler when some errors occur to zonex, so hvisor doesn't need to panic. But due to the bug with shutdown(#131 ), it is useless now...


src/hypercall/mod.rs
```diff
@@ -275,17 +284,28 @@ impl<'a> HyperCall<'a> {
...
+        let mut count: u32 = 0;
+
         // wait all zone's cpus shutdown
         while zone_r.cpu_set.iter().any(|cpu_id| {
             let _lock = get_cpu_data(cpu_id).ctrl_lock.lock();
-            get_cpu_data(cpu_id).arch_cpu.power_on
+            let power_on = get_cpu_data(cpu_id).arch_cpu.power_on;
+            count += 1;
+            if count > 10000000 {
+                if(power_on) {
```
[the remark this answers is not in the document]
Inquisitor-201: Maybe we should rewrite **platform/aarch64/qemu-gicv3/test/systemtest/testcase/tc_zone_list2.txt** as below?
|     zone_id     |       cpus        |      name       |     status |
|               0 |              0, 1 |      root-linux |    running |
|               1 |              2, 3 |          linux2 |    running |

KouweiLee:
Not only this txt, the rootfs should also be changed. I’m waiting chh to do these




---- Replied Message ----
| From | ***@***.***> |
| Date | 04/06/2025 21:00 |
| To | syswonder/hvisor ***@***.***> |
| Cc | KouweiLee ***@***.***>,
Author ***@***.***> |
| Subject | Re: [syswonder/hvisor] add config check and add zone error handler to avoid panic (PR #132) |

Maybe we should rewrite platform/aarch64/qemu-gicv3/test/systemtest/testcase/tc_zone_list2.txt like the attachment file?
tc_zone_list2.txt

—
Reply to this email directly, view it on GitHub, or unsubscribe.
You are receiving this because you authored the thread.Message ID: ***@***.***>

Inquisitor-201 left a comment (syswonder/hvisor#132)

Maybe we should rewrite platform/aarch64/qemu-gicv3/test/systemtest/testcase/tc_zone_list2.txt like the attachment file?
tc_zone_list2.txt

caodg:
> difference
maybe provide two shutdown solutions, one is graceful, waiting all events finish, one is brute and immediately, as your current solution. e.g., shutdown vs  shutdown -f

KouweiLee:
> maybe provide two shutdown solutions, one is graceful, waiting all events finish, one is brute and immediately, as your current solution. e.g., shutdown vs shutdown -f, and administrator makes decision.

In Arm GIC, if an interrupt is triggered many times before hvisor reacts to it, hvisor will only react once. When hvisor reacts to the interrupt, it will only handle one event. So there are events remaining to be handled. Unless hvisor handles all events once it gets an sgi interrupt, it can't wait all events finished. But according to @enkerewpo, handling all events in one sgi interrupt is not safe and may cause more bugs. So I think it's not the problem of shutdown, it's about the event handle program.

src/hypercall/mod.rs
Inquisitor-201, line 1: The condition count > 10000000 is not very standardized. Consider replacing it with a fixed constant (e.g., SHUTDOWN_MAX_COUNTER) for clarity and maintainability.

src/zone.rs
Inquisitor-201, line 1: Can the **error!** macro and the **zone_error** function be combined into a **zone_error!** macro?

src/hypercall/mod.rs
Inquisitor-201, line 1: Under what situations would shutting down CPUs in non-root zones fail?

KouweiLee: > Under what situations would shutting down CPUs in non-root zones fail?
When zone_error!() is occured, users should shutdown cpus. What do you mean?

KouweiLee:
I have said before, when non root uses virtio devices, the rcu would occur. But I have fixed it, so no rcu problem anymore 


---- Replied Message ----
| From | ***@***.***> |
| Date | 04/08/2025 11:40 |
| To | syswonder/hvisor ***@***.***> |
| Cc | KouweiLee ***@***.***>,
Author ***@***.***> |
| Subject | Re: [syswonder/hvisor] add config check and add zone error handler to avoid panic (PR #132) |

@Inquisitor-201 commented on this pull request.

On src/hypercall/mod.rs:

Under what situations would shutting down CPUs in non-root zones fail?
When zone_error!() is occured, users should shutdown cpus. What do you mean?

For example, when the hvisor-tool sends a shutdown request, the hvisor will keep waiting(leading to a RCU error) until all CPUs in non-root are shut down. Under what circumstances would other CPUs fail to shut down, causing the hvisor to enter a waiting state?

Inquisitor-201 (approved):

PR #133, opened 2025-04-05
feat: Enable dynamic CPU detection for ARM platforms via device tree

FlowerBlackG:
Previously, the system was limited to a maximum of 4 CPUs. This change allows the system to dynamically detect the number of available CPUs on ARM platforms by reading the device tree and utilize them fully (up to 32 cores).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -98,8 +98,36 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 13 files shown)


KouweiLee:
You should change your pr destination to branch `dev`, not `main`

PR #137, opened 2025-04-09
riscv64: add virtio support

liulog:
- modify zone0's cpu num to 2
- add HW_IRQS in board.rs (only these irqs will be operated in related physical PLIC)
- virtio irq is 0x20 (different from arm's 0x20+32)
- add related json files in platform dir ( can easily repeat )

Below is the output of zone1 with virtio console:
image

### DIFF
--- a/platform/riscv64/qemu-plic/board.rs
+++ b/platform/riscv64/qemu-plic/board.rs
@@ -26,7 +26,7 @@ pub const BOARD_PLIC_INTERRUPTS_NUM: usize = 1023;  // except irq 0
 pub const ROOT_ZONE_DTB_ADDR: u64 = 0x8f000000;
 pub const ROOT_ZONE_KERNEL_ADDR: u64 = 0x90000000;
 pub const ROOT_ZONE_ENTRY: u64 = 0x90000000;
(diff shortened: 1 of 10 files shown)


liulog:
@SanchezDorso please check it and review. 

Thanks.

liulog:
You can use the commands below:

```bash
# print zone0 cpuinfo
mount -t proc proc /proc
mount -t sysfs sysfs /sys
cat /proc/cpuinfo

cd /root
insmod ./hvisor.ko
rm nohup.out
mkdir -p /dev/pts
mount -t devpts devpts /dev/pts
nohup ./hvisor virtio start virtio_cfg.json &
./hvisor zone start zone1-linux-virtio.json && \
cat nohup.out | grep "char device" && \
script /dev/null

# link to virtio-console, then goto zone1's virtio-console 
screen /dev/pts/0

# print zone1's cpuinfo
mount -t proc proc /proc
mount -t sysfs sysfs /sys
cat /proc/cpuinfo
```

SanchezDorso (commented):
Have checked it and I can also run zone1 with virtio console by using hvisor-tool commit 3914f6b56afcc4645b2bb2292ba1dfc685a381c2.

enkerewpo (approved):

PR #142, opened 2025-04-13
feat: add the CI workflow to support running CI with the latest hvisor-tool and the configuration files in hvisor.

CHonghaohao:
1. Pull the latest hvisor-tool in the workflow and compile it. Finally, place the kernel module and command-line tool into the rootfs.
2. When starting zone1 in the workflow, use the configuration files and scripts in the hvisor directory.
3. The above tasks can dynamically modify the CI workflow without re - uploading the rootfs.
4. Resume the test to shut down zone1


KouweiLee (approved):
A very good job!

PR #144, opened 2025-04-15
riscv64: fix zone shutdown bug

liulog:
- modify riscv64 ArchCpu's run and idle function
- support riscv64's arch_irqchip_reset (clear related PLIC)
- add clear_events, when zone is shutdown, we should clear events related to this cpu


liulog:
Below is the example:
image

enkerewpo (approved):

PR #160, opened 2025-06-07
riscv64: support unittest

liulog:
I have ported unittest to riscv64.
Use `make BID=riscv64/qemu-plic test`, (platform qemu-aia also works) then you can get this:
image

### DIFF
--- a/platform/riscv64/qemu-aia/test/runner.sh
+++ b/platform/riscv64/qemu-aia/test/runner.sh
@@ -1,0 +1,98 @@
+#!/bin/bash
+
+set -e
+
+PWD=$(pwd)
+THIS=$(basename $0)
+CARGO_BUILD_INPUT_ARG0=$1
+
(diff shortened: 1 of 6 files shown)


platform/riscv64/qemu-aia/test/runner.sh
enkerewpo, line 1: since the test script is exclusive for each board after the platform abstraction, you can just leave necessary commands for riscv and this specific board, and remove aarch64 stuff

platform/riscv64/qemu-plic/test/runner.sh
enkerewpo, line 1: similarly you can just write riscv commands here

liulog: Yes, you're right. I have updated shell script about riscv64 as you said above.

enkerewpo (approved):

PR #166, opened 2025-06-18
feat(loongarch). add pci passthrough/allocation support for 3A5000/6000

enkerewpo:
note:
1. MAX PCI DEVICE number set to `32` by Zhongkai Xu.
2. MAX VIRTIO DEVICE number set to `8` because we have to boot 6 virtio devices on 3A6000 for 3 nonroots.
3. currently the `ls3a5000` platform config can be used for both 3A5000 and 3A6000 (+7A2000) because we didn't add support for SMT on 3A6000 so everything else is the same.

tested on official TIMCREATE 3A6000 board:
1. root boot ok
4. nonroot boot ok
5. root and nonroot pci allocation and virtualization ok
6. nonroot virtio-console and virtio-blk ok.

todo: fix problems shown in CI and test on official TIMCREATE 3A5000 board.
upd: tested


enkerewpo:
UPD: tested on 3A5000 and everything is ok.
weird thing is that in root linux on 3A5000, `sleep 1` seems ok even with all eiointc stuff, but nonroot's `sleep 1` is still wrong(2s)

PR #173, opened 2025-07-02
[feature] riscv64: add virtio support in qemu-aia to boot zone1.

CHonghaohao:
- add interrupt injection support for aia
- add handling for virtio mmio page faults
- add configuration to boot zone1 via virtio devices


liulog:
I think there is room for improvement in AIA's code. 
The current modifications have a lot of redundancy, and it would be more appropriate to merge this part after modifications with AIA.

CHonghaohao:
> I think there is room for improvement in AIA's code. The current modifications have a lot of redundancy, and it would be more appropriate to merge this part after modifications with AIA.

There is indeed much room for improvement in AIA's code, and it can be upgraded later. However, regarding the newly added features here, can we first merge them?

Inquisitor-201 (approved):

PR #178, opened 2025-07-15
riscv64: initial support milk-v megrez board

liulog:
Now, we can passthrough blk and uart to zone0 and zone1. 
(virtio console and virtio-blk have also been supported.)
We assigned uart0(ttyS0) and mmc(using SD card) to zone0, then assigned uart2(ttyS2) and sata with m.2 key to zone1.
The result is below:
a23aa8e3340283b23762c47f404d690

Note: current both zone0 and zone1 need cache-controller and syscon (you can get related configuration in megrez's config dir). They are needed for driving dma-noncoherent devices(such as mmc and sata in megerz). If zone1 is configured with virtio-blk, this is not necessary.


dallasxy (changes requested):
Please avoid excessive use of warn logs in non-development environments. info might be more appropriate.

PR #180, opened 2025-07-18
riscv64: mv clear_bss to entry assembly code Fix #179

### LINKED ISSUE #179: riscv64: in physical board, sometimes hvisor stalls in println
When I'm porting hvisor to milk-v megrez & sifive hifive-premier-p550 both which use eic7700x soc, I found that sometimes hvisor stalls in println!.

Sometimes, I could only see "Starting kernel" in serial monitor(Mobaxterm), hvisor doesn't have any log message outputted.

Futhermore, I found hvisor stall in PRINT_LOCK in function print. So I suspect that the PRINT_LOCK's initial value is error.

Through scanning the entry code, I found clear_bss is placed after the first println, this may be root reason for the problem. 

PRINT_LOCK's content would be placed in bss segment by compiler. It is needed to use PRINT_LOCK after clearing bsss.

I'm fixing it, see one Pull-Request soon!

liulog:
This PR is to fix #179.

Mv clear_bss function before println, because println uses PRINT_LOCK which is in bss segment.

I use cpu0 to clear_bss, it could see cleared bss segment sure.

In order to let other cores could see cleared bss, I use `fence rw, rw` to ensure other cores run into rust_main after seeing cleared bss segment.

### DIFF
--- a/src/arch/riscv64/entry.rs
+++ b/src/arch/riscv64/entry.rs
@@ -17,7 +17,11 @@ use crate::consts::PER_CPU_SIZE;
 
 #[no_mangle]
 #[link_section = ".data"]
-pub static mut CPU0_BSS_LOCK: u32 = 1;
+pub static mut CPU_BSS_LOCK: u32 = 1;
+
+#[no_mangle]
+#[link_section = ".data"]
+pub static mut ENTER_CPU: u32 = u32::MAX; // the first entered cpuid will be written.
 
 #[naked]
 #[no_mangle]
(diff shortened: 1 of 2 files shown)


liulog:
@CHonghaohao  Please review it.

Now it is needed for eic7700x soc.

liulog:
Before, I want use cpu0 to clear bss segment, but I forget the first entry cpu could not be cpu0.
So I later modify assembly code, using the first entry cpu to clear bss segment.

PR #183, opened 2025-07-19
Code simplification CI testing

ForeverYolo:
This PR is currently used through CI to test the correctness of code simplification.


src/arch/aarch64/cpu.rs
```diff
@@ -215,13 +218,26 @@ impl ArchCpu {
...
-    }
+    let mpidr = mpidr & MPIDR_MASK;
+    (0..MAX_CPU_NUM)
+        .find(|&i| BOARD_MPIDR_MAPPINGS[i] == mpidr)
+        .unwrap() as u64
+}
+
+pub fn cpuid_to_mpidr_affinity(cpuid: u64) -> (u64, u64, u64, u64) {
+    let mpidr = BOARD_MPIDR_MAPPINGS[cpuid as usize];
+    let aff3 = (mpidr >> 24) & 0xff;
```
dallasxy, line 229: The mask is 0xff00ffffff, but bit 24-31 is used as aff3. In fact, I looked up the manual and found that the correct mask for MPIDR_EL1 should be 0xff41ffffff, and aff3 here needs to be modified to (mpidr >> 32) & 0xff

dallasxy (changes requested):

Inquisitor-201 (approved):

dallasxy (approved):

PR #186, opened 2025-07-26
refactor: move plic from zone struct

liulog:
Now there is no cfg about feature plic in arch dir.


liulog:
I add one function for zone, named "virq_init".
Now this is only used for feature plic. @Inquisitor-201

PR #193, opened 2025-09-06
Merge new features to main for hvisor-v0.2 release

Inquisitor-201:


src/arch/aarch64/cpu.rs
```diff
@@ -215,13 +245,26 @@ impl ArchCpu {
...
-    }
+    let mpidr = mpidr & MPIDR_MASK;
+    (0..MAX_CPU_NUM)
+        .find(|&i| BOARD_MPIDR_MAPPINGS[i] == mpidr)
+        .unwrap() as u64
+}
+
+pub fn cpuid_to_mpidr_affinity(cpuid: u64) -> (u64, u64, u64, u64) {
+    let mpidr = BOARD_MPIDR_MAPPINGS[cpuid as usize];
+    let aff3 = (mpidr >> 32) & 0xff;
```
KouweiLee, line 256: This may be `aff3 = (mpidr >> 24)` ?

src/device/irqchip/gicv3/mod.rs
```diff
@@ -345,11 +346,24 @@ pub fn host_gicd_base() -> usize {
...
-}
 
+    let mpidr = BOARD_MPIDR_MAPPINGS[id];
+
+    let order = SORTED_MPIDRS
+        .iter()
+        .position(|&x| x == mpidr)
+        .expect("MPIDR from BOARD_MPIDR_MAPPINGS must be present in sorted list");
+
+    GIC.get().unwrap().gicr_base + order * PER_GICR_SIZE
```
KouweiLee, line 365: I have a question here. The correct way to find the gicr base of one cpu is to check if `GICR_TYPER.Affinity_value` is equal to the mpidr of this cpu. I searched in the internet and found that the order of the cpu's mpidr may not be the order of the cpu's gicr_base.

KouweiLee (changes requested):

src/arch/loongarch64/hypercall.rs
```diff
@@ -0,0 +1,69 @@
...
+
+use crate::arch::cpu::this_cpu_id;
+use crate::config::HvZoneConfig;
+use crate::config::CONFIG_MAGIC_VERSION;
+use crate::device::virtio_trampoline::MAX_DEVS;
+use crate::hypercall::HyperCall;
+use crate::hypercall::HyperCallResult;
+impl<'a> HyperCall<'a> {
+    pub fn hv_ivc_info(&mut self, ivc_info_ipa: u64) -> HyperCallResult {
+        warn!("hv_ivc_info is not implemented for Risc-V");
```
enkerewpo, line 25: `Risc-V` should be `LoongArch64` ?

caodg (commented):
we should indicate the current status, the main and new features,  in README, and list major changes in CHANGELOG.md

src/arch/aarch64/cpu.rs
Inquisitor-201: image
According to [https://developer.arm.com/documentation/ddi0601/2025-06/AArch64-Registers/MPIDR-EL1--Multiprocessor-Affinity-Register](url), `aff3` should be `mpidr >> 32`

KouweiLee: It is reasonable. All right.

PR #204, opened 2025-10-18
x86_64: Add support for QEMU Q35 and ASUS NUC14MNK

Solicey:


src/device/uart/mod.rs
```diff
@@ -46,3 +46,10 @@ pub use loongson_uart::{console_getchar, console_putchar};
 mod uart_16550;
 #[cfg(all(feature = "uart_16550", target_arch = "aarch64"))]
 pub use uart_16550::{console_getchar, console_putchar};
+
+#[cfg(target_arch = "x86_64")]
```
Inquisitor-201: You can add a new cargo feature here, like #[cfg(all(feature = "uart16550a", target_arch = "x86"))]

src/hypercall/mod.rs
```diff
@@ -37,6 +38,7 @@ numeric_enum! {
     pub enum HyperCallCode {
         HvVirtioInit = 0,
         HvVirtioInjectIrq = 1,
+        HvVirtioGetIrq = 86,
```
Inquisitor-201: Why set this hypercall code to 86?

src/logging.rs
```diff
@@ -137,6 +159,19 @@ impl Log for SimpleLogger {
             Level::Debug => ColorCode::Cyan,
             Level::Trace => ColorCode::BrightBlack,
         };
+
+        #[cfg(all(feature = "graphics"))]
```
Inquisitor-201: You can put the `graphics` logic in a separate function to hide the details.

src/arch/aarch64/cpu.rs
```diff
@@ -268,3 +269,12 @@ pub fn store_cpu_pointer_to_reg(pointer: usize) {
     // println!("aarch64 doesn't support store cpu pointer to reg, pointer: {:#x}", pointer);
     return;
 }
+
+pub fn get_target_cpu(irq: usize, zone_id: usize) -> usize {
```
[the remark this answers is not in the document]
Solicey: This is because in x86_64, the target CPU of virtio device interrupt is decided by virtual I/O APIC, while in other archs we can simply use the first CPU as the interrupt receiver. Therefore, I add a generic API `get_target_cpu` for all archs to decide which CPU should we set as target.

src/device/uart/mod.rs
Solicey: Sure, I will add an "uart16550a" feature in Cargo.toml

PR #238, opened 2026-01-07
add percpu support

li041:
- This change has not been fully tested yet. Currently it has only been verified on qemu-gicv3, qemu-aia, and rk3588.
- I would appreciate it if other developers could help test this on additional platforms.
- The main purpose of this PR is to introduce the percpu crate to better abstract per-CPU variables, instead of relying on global Vec.
- Feedback on both correctness and design is very welcome.

### DIFF
--- a/platform/aarch64/qemu-gicv3/linker.ld
+++ b/platform/aarch64/qemu-gicv3/linker.ld
@@ -38,7 +38,7 @@ SECTIONS
         *(.sbss .sbss.*)
     }
 
-   . = ALIGN(4K);
+    . = ALIGN(4K);
     _percpu_start = .;
     _percpu_end = _percpu_start + SIZEOF(.percpu);
     .percpu 0x0 (NOLOAD) : AT(_percpu_start) {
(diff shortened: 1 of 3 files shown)


ForeverYolo:
Can you explain why we need to do this?

li041:
> Can you explain why we need to do this?

less code, unified per-CPU abstraction, possible performance improvement

enkerewpo (approved):

platform/x86_64/qemu/linker.ld
```diff
@@ -41,6 +42,17 @@ SECTIONS
...
     }
 
+    . = ALIGN(4K);
+    _percpu_start = .;
+    _percpu_end = _percpu_start + SIZEOF(.percpu);
+    .percpu 0x0 (NOLOAD) : AT(_percpu_start) {
+        _percpu_load_start = .;
+        *(.percpu .percpu.*)
+        _percpu_load_end = .;
+        . = _percpu_load_start + ALIGN(64) * CPU_NUM;
```
Solicey: Why are we using ALIGN(64) here? By the way, could you explain the purpose of reserving the `.percpu` segment?

Solicey (approved):
works fine on x86_64 QEMU

PR #245, opened 2026-01-21
fix(Makefile): fixed memory size calculation on loongarch64

### LINKED ISSUE #244: bug: on loongarch64, `MEM_SIZE` in build summary is not correct
Image

enkerewpo:
fixed #244 

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,7 +91,7 @@ all: clean_check gen_cargo_config $(hvisor_bin)
 	@printf "%-10s %s\n" "HVISON_BIN_SIZE =" "$(COLOR_BOLD)$(shell du -h $(hvisor_bin) | cut -f1)$(COLOR_RESET)"
 	@start_addr=$$(rust-nm $(hvisor_elf) | grep skernel | awk '{print $$1}'); \
 	 end_addr=$$(rust-nm $(hvisor_elf) | grep __hv_end | awk '{print $$1}'); \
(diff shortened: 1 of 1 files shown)


agicy:
Thanks for the PR.

The current `$((0x...))` arithmetic relies on 64-bit signed integers, which can indeed overflow when dealing with high-bit addresses on certain platforms. Unfortunately, due to the lack of a local LoongArch environment, I wasn't able to catch this during initial testing.

However, switching to `awk` with `strtonum` introduces two risks:
1. **Portability**: `strtonum` is a GNU-specific extension and is not part of POSIX `awk`.
2. **Precision**: `awk` typically uses double-precision floating-point numbers (IEEE 754), which only provide 53 bits of precision. This may cause rounding errors for 64-bit addresses.

I suggest using `bc` instead. It is POSIX-compliant, handles arbitrary precision, and is more robust:

```
size=$$(echo "obase=16; ibase=16; $$(echo $$end_addr | tr 'a-z' 'A-Z') - $$(echo $$start_addr | tr 'a-z' 'A-Z')" | bc | tr 'A-Z' 'a-z');
```
(Note: `bc` requires uppercase hex input, so `tr` is used.)

What do you think? Could you please verify if this `bc` approach works in your LoongArch environment?

enkerewpo:
I’ve updated the commit and tested the `bc` method on LoongArch64 and it works as expected.

PR #260, opened 2026-02-25
Bugfix: Pin cargo-xbuild to v0.6.5 in CI

Inquisitor-201:
The CI pipeline recently started failing because cargo-xbuild v0.6.6 (and its dependencies) requires edition2024, which is not supported by our current nightly Rust toolchain (1.80.0).

This PR locks the version to v0.6.5 to restore CI stability while keeping the existing toolchain environment.

Related Error:
feature edition2024 is required but not stabilized in this version of Cargo

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,6 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y qemu-system-aarch64 qemu-system-riscv64 gdb-multiarch llvm-dev libclang-dev wget
           cargo install --version 0.3.0 --locked cargo-binutils
-          cargo install --version 0.6.5 --locked cargo-xbuild
       - name: Format Check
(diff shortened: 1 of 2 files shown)


li041:
Maybe `cargo install cargo-xbuild` is redundant. The built-in `-Zbuild-std` feature in Cargo's nightly channel now handles cross-compiling the sysroot (core, alloc, std) automatically, and I do not see any usage of `cargo-xbuild` in our repo.

dallasxy (approved):

Inquisitor-201:
> Maybe `cargo install cargo-xbuild` is redundant. The built-in `-Zbuild-std` feature in Cargo's nightly channel now handles cross-compiling the sysroot (core, alloc, std) automatically, and I do not see any usage of `cargo-xbuild` in our repo.

Thanks, I'll see if it works.

Issue #239, opened 2026-01-09
RT-Thread guest: serial console stuck (no output) after shutdown + switching assigned CPU core (RK3588)


Jaxtonmax:
### Description
When running an RT-Thread guest on RK3588 via hvisor, the guest boots and prints UART logs normally on the first run. However, if we shutdown the RT-Thread guest, change the zone config CPU assignment (e.g. cpus: [7] → cpus: [6]), and start the same guest again, the serial console becomes completely silent (no output). This issue is reproducible and looks correlated with switching the assigned physical CPU core after a prior shutdown.

### How to reproduce:

1. Boot Root Linux on RK3588 with hvisor.ko loaded.
2. Start an RT-Thread guest .Initial config uses cpus: [7] (works).
3. On Root Linux, shutdown the RT-Thread zone:

-     hvisor zone shutdown -id <zone_id>

4. Modify the RT-Thread zone config to use a different CPU core, e.g.:cpus: [7] → cpus: [6] 
5. Start the same RT-Thread zone again.
6. Observe the serial output.

### Behaviour

- Expected:RT-Thread should boot normally and print logs to UART after restart, regardless of which allowed pCPU core it is bound to.

- Actual:After shutdown + switching cpus to another core, the UART console shows a black screen (no output).

### Environment

- Architecture: aarch64
- platform: rk3588

liulog:
The reason is that the memory that has been used before is dirty.

Before running a vm, hvisor&hvisor-tool should confirm that the memory is clean.

liulog:
https://github.com/syswonder/hvisor/pull/263
https://github.com/syswonder/hvisor-tool/pull/79

I used a script to run the restart process 50 times, randomly switching the core, and it worked perfectly. The serial port was interactive.

<img width="1164" height="768" alt="Image" src="https://github.com/user-attachments/assets/31cbccc2-f63d-4500-9b02-1b461a2d5141" />

Jaxtonmax:
I tested this on RK3588 after applying the latest hvisor and hvisor-tool changes.

Test scope:
- RT-Thread guest restart after shutdown + switching assigned CPU core
- Linux guest restart after shutdown + switching assigned CPU core
- Repeated zone start/shutdown with different CPU assignments

Test method:
- Replaced the old setup with the latest hvisor changes and the latest hvisor-tool API path
- Rebuilt and deployed fitImage, hvisor.ko and the hvisor userspace tool
[...]

PR #263, opened 2026-03-03
enhance: add dcache invalid ops for aarch64

liulog:
A flush operation for dcache and a TBLI operation for EL1 were added, and then Icache was cleaned up to ensure that there was no dirty data in the cache when the guest was first executed.

Related with https://github.com/syswonder/hvisor-tool/pull/79 and https://github.com/syswonder/hvisor/issues/239

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -267,7 +267,7 @@ checksum = "8a9f4cc54a2e471ff72f1499461ba381ad4eae9cbd60d29c258545b995e406e0"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.111",
 ]
 
 [[package]]
--- a/src/arch/aarch64/cache.rs
+++ b/src/arch/aarch64/cache.rs
@@ -11,7 +11,8 @@
 // Syswonder Website:
 //      https://www.syswonder.org
 //
-// Authors:
+// Authors: 
+//  Jingyu Liu <liujingyu24s@ict.ac.cn>
 //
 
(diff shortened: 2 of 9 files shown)


liulog:
Test on RK3588 (switch rt-thread to 2,3,4 is okay) has been done.

Verification by others is required.

ForeverYolo:
A question: if Linux is booted, before executing the first line of Linux assembly, the Cache and MMU are in a disabled state, and the cleanup work is also done by Linux. What is the purpose of doing it this way? @liulog

liulog:
> A question: if Linux is booted, before executing the first line of Linux assembly, the Cache and MMU are in a disabled state, and the cleanup work is also done by Linux. What is the purpose of doing it this way? @liulog


The issue here concerns keeping the VM image in memory clean before it starts, allocating RAM, associated CPU cache, and TLB to the VM.

By default, the VM reads the image from memory early in the startup process. However, in special cases, the image may be in the root zone cpu's  private cache. Additionally, if the cache and TLB are not clean, dirty data may accumulate in them during zone restarts, etc. When the VM has cache and MMU translation enabled, it may read previously uncleaned dirty data from the cache and TLB, causing errors.

This is just a precaution; it cannot be guaranteed that all operating systems will clear caches and TLB in the early stages of startup.

Jaxtonmax:
The issue has been resolved on RK3588. Here is my detailed test report:https://github.com/syswonder/hvisor/issues/239#issuecomment-4072623221

ForeverYolo (approved):

Issue #261, opened 2026-02-26
aarch64: restart zone1-linux failed due to gits


liulog:
Restarting zone1-linux failed. After tracing with gdb, it was found that the problem was caused by a loop in the hvisor at this point.

<img width="1280" height="763" alt="Image" src="https://github.com/user-attachments/assets/b9ba0c03-f449-463f-8d83-3c2890e48c08" />

I think some content in gits needs to be reset.

ZhongkaiXu:
Thanks for the report.

The reason for this error is that after a restart, Linux resets the `ITS_WRITER`  to 0, but hvisor still keeps the virtual `ITS_READR` from the previous lifecycle. Since the CMD ring allows writer < readr to indicate wrap-around usage, this results in meaningless CMD writes and the loop cannot finish cause hardware find the cmds software writing are wrong cmds.

I will add logic during shutdown to clear the old ITS-related virtualization state.

PR #264, opened 2026-03-03
Fix PCIe BAR mapping bug and add gits-reset

ZhongkaiXu:
**PCI BAR GPA-HPA mapping bug** : In the old code, during BAR GPA-to-HPA mapping, the HPA value was incorrectly overwritten with the GPA value. This caused guest accesses to be routed to an unknown address range, making the device unusable.

#261 The reason for this error is that after a restart, Linux resets the ITS WRITER register to 0, but hvisor still keeps the virtual READAR from the previous lifecycle. Since the CMD ring allows WRITER< READR to indicate wrap-around usage, this results in meaningless CMD writes and the loop cannot finish.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -267,7 +267,7 @@ checksum = "8a9f4cc54a2e471ff72f1499461ba381ad4eae9cbd60d29c258545b995e406e0"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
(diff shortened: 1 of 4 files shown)


dallasxy:
The fix for the PCIe BAR has been tested and is working correctly.

dallasxy (approved):

PR #270, opened 2026-03-16
Separate lock-free fields from `Zone` and make fields private

li041:
## Overview

This PR refactors the `Zone` structure to clearly separate fields that require synchronization from fields that do not, and improves encapsulation by removing direct public field access.

### 1. Separate lock-free fields

Previously, all fields were stored directly inside `Zone`, even though some fields do not require synchronization.

This PR introduces a new structure:

```rust
pub struct ZoneInner
```

which contains fields that require synchronization. `ZoneInner` is protected by an `RwLock` inside `Zone`.

Fields that do not require whole locking remain directly in `Zone`, such as:

* `name`
* `id`
* `is_err`

This separation makes it explicit which fields require locking and avoids unnecessary locking for immutable or atomic data.

### 2. Make struct fields private

All fields in `Zone` and `ZoneInner` were previously declared as `pub`.
They are now made **private** and accessed through accessor methods.

This prevents external code from directly mutating internal state and ensures that access patterns follow the intended synchronization model.

## Result

After this change:

* Lock-free fields are clearly separated from synchronized state.
* Access to internal fields is controlled through APIs.

### DIFF
--- a/src/zone.rs
+++ b/src/zone.rs
@@ -18,7 +18,7 @@ use alloc::vec::Vec;
 // use psci::error::INVALID_ADDRESS;
 use crate::consts::{INVALID_ADDRESS, MAX_CPU_NUM};
 use crate::pci::pci_struct::VirtualRootComplex;
-use spin::RwLock;
+use spin::{RwLock, RwLockReadGuard, RwLockWriteGuard};
 
 #[cfg(feature = "dwc_pcie")]
 use crate::pci::{config_accessors::dwc_atu::AtuConfig, PciConfigAddress};
@@ -34,6 +34,7 @@ use crate::error::HvResult;
 use crate::memory::addr::GuestPhysAddr;
 use crate::memory::{MMIOConfig, MMIOHandler, MMIORegion, MemorySet};
 use core::panic;
+use core::sync::atomic::{AtomicBool, Ordering};
 
 #[cfg(feature = "dwc_pcie")]
 #[derive(Debug)]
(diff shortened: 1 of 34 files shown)


src/zone.rs
```diff
@@ -108,36 +109,77 @@ impl VirtualAtuConfigs {
...
-    pub gpm: MemorySet<Stage2PageTable>,
-    pub iommu_pt: Option<MemorySet<Stage2PageTable>>,
-    pub is_err: bool,
-    pub vpci_bus: VirtualRootComplex,
+    name: [u8; CONFIG_NAME_MAXLEN],
+    id: usize,
+    is_err: AtomicBool,
+    inner: RwLock<ZoneInner>,
+}
+
```
liulog, line 117: Some feilds in `Zoneinner` could also be moved to `Zone`, For examples, cpu_num, cpu_set, irq_bitmap. So I think we should discuss this before merge.

li041: Thanks a lot for bringing this up! You're absolutely right, fields like `cpu_num`, `cpu_set`, and `irq_bitmap` are indeed good candidates for discussion regarding their placement between `Zone` and `ZoneInner`.

I agree that, looking ahead, these fields might need to be modified in the future. The decision of where to place them should balance current logic convenience with **future maintainability and clarity**.

Besides, we currently initialize the Zone first, then read the corresponding configuration to set these values. This means these fields need **internal mutability** - they are written after the main zone structure is created. If we move them to Zone (the outer struct), we might break this initialization pattern or require changes to how we protect these writes. By keeping them in ZoneInner, we can maintain the existing locking semantics - modifications can be done while holding the appropriate lock (e.g., the zone's inner lock), without changing the lock protection strategy we already have in place.

liulog (approved):

ForeverYolo:
Has this PR been tested on any development board so far?

ForeverYolo:
@li041

li041:
> Has this PR been tested on any development board so far?

Yeah, this PR has been tested on qemu platform and Rk3588.

PR #274, opened 2026-03-19
Rollback partial page allocation on failure

### LINKED ISSUE #268: Partial page table mappings when allocator runs out of frames during region insertion
When inserting a new memory region into a MemorySet, the mapping procedure may fail after partially mapping the region if the physical frame allocator runs out of free frames.

The current workflow is:
```
/// Add a memory region to this set.
pub fn insert(&mut self, region: MemoryRegion<PT::VA>) -> HvResult {
    ...
    self.pt.map(&region)?;
    self.regions.insert(region.start, region);
    Ok(())
}
```

During map():

```
fn map(&mut self, region: &MemoryRegion<VA>) -> HvResult {
    ...
    while size > 0 {
        ...
        self.inner
            .map_page(page, paddr, region.flags)
            .map_err(|e: PagingError| {
                error!(
                    "failed to map page: {:#x?}({:?}) -> {:#x?}, {:?}",
                    vaddr, page_size, paddr, e
                );
                e
            })?;
        ...
    }
    Ok(())
}
```

If the allocator cannot provide enough frames during page table creation (e.g., intermediate page tables), map_page() returns an error and the mapping process aborts.

**Problem**
This leads to an inconsistent state:
- Some virtual pages in the region are mapped
- Others remain unmapped
- The region is not inserted into `self.regions`
- Page table state and memory region metadata become inconsistent

li041:
Close #268
## Summary
- Implement rollback for partial page allocation failures 

## Motivation
Partial allocation failures previously risked leaving the allocator in an inconsistent state, causing leaks or corruption. Rolling back partial allocations makes failure modes safe and observable without aborting execution.

### DIFF
--- a/src/arch/aarch64/ivc.rs
+++ b/src/arch/aarch64/ivc.rs
@@ -150,7 +150,7 @@ impl From<&HvIvcConfig> for IvcRecord {
 }
 
 impl Zone {
-    pub fn ivc_init(&mut self, ivc_configs: &[HvIvcConfig]) {
+    pub fn ivc_init(&mut self, ivc_configs: &[HvIvcConfig]) -> HvResult {
         let mut inner = self.write();
         for ivc_config in ivc_configs {
             // is_new is ok to remove
@@ -170,8 +170,7 @@ impl Zone {
                         start_paddr,
                         rw_sec_size as _,
                         MemFlags::READ | MemFlags::WRITE,
-                    ))
-                    .unwrap();
+                    ))?;
                 for i in 0..ivc_config.max_peers as usize {
                     let flags = if i == ivc_config.peer_id as _ {
                         MemFlags::READ | MemFlags::WRITE
@@ -185,8 +184,7 @@ impl Zone {
                             start_paddr + rw_sec_size + i * out_sec_size,
                             out_sec_size as _,
                             flags,
-                        ))
-                        .unwrap();
+                        ))?;
                 }
                 inner.mmio_region_register(
                     ivc_config.control_table_ipa as _,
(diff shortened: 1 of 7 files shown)


liulog:
Perhaps we need more logic? 
For example, if we fail to map all memoryset required by the zonex, `zone create` should return fail.

li041:
> Perhaps we need more logic? For example, if we fail to map all memoryset required by the zonex, `zone create` should return fail.

Yeah, that's current error handling. [zone_create](https://github.com/syswonder/hvisor/blob/341a394b3672e3627dbe8329a7b8082f5d4189a8/src/zone.rs#L375) may  panic in the context of `ivc_init` before, and I've fixed it in this commit. `zone_create` return a `Result`, and it let the caller to decide the action when encountering map error. In details, zone0 call `zone_create` with `unwarp` in  `primary_init_early` while `zoneU`  is created by `hv_zone_start` which just pass the error to the upper caller.

liulog:
It addresses the issue of incomplete mapping in a specific region. Therefore, it mainly targets rollbacks where a region mapping is incomplete.
For the remaining fully mapped portions (in `zone.gpm`) of the entire memoryset, RAII will automatically perform the unmapping process.

liulog (approved):

ForeverYolo:
Any test on board?

ForeverYolo:
@li041

li041:
> Any test on board?

I haven't tested this on a physical board yet. However, I have added comprehensive error handling to the implementation and have specifically constructed error scenarios in QEMU to verify these paths. Given that the logic has been validated through these simulated fault injections, I believe the current testing is sufficient for this PR.

PR #295, opened 2026-04-10
iommu: code refactor, decouple from arch

### LINKED ISSUE #292: Refactor the iommu part of the code, decoupling the implementation from arch.
Currently, the IOMMU code is embedded under the arch directory, which results in high coupling and hinders expansion. 

We are considering moving it to the same directory as irqchip and providing a trait to offer a unified interface.

liulog:
## Purpose

Close #292 Refactor the iommu part of the code, decoupling the implementation from arch.


liulog:
The structure of the IOMMU subsystem is shown in the figure below. To maintain compatibility with the original implementation, an encapsulation layer has been added in the middle.

image

Solicey (approved):

PR #303, opened 2026-05-09
feat: extend vcpu state

liulog:
Expand the vCPU states to allow for additional suspend states.

The power_on state overlaps with the vCPU state; remove power_on and replace it with the vCPU state.

Currently, only these four states are used; more will be supported in the future.

```rust
/// VCpu lifecycle states
#[repr(u8)]
#[derive(Copy, Clone, Debug, Eq, PartialEq)]
pub enum VcpuState {
    /// Initial state or after PSCI CPU_OFF. Not in any run queue.
    Stopped = 0,
    /// In a pCPU's run queue, waiting to be scheduled.
    Ready = 1,
    /// Currently executing on a pCPU.
    Running = 2,
    /// Blocked by WFI/CPU_SUSPEND. Not in any run queue, awaiting interrupt wakeup.
    Blocked = 3,
}
```

Architecture-independent, it's placed under percpu. In the future, when vcpu support is added, the percpu content will need to be refactored, and the vcpu-related parts will be separated.

```rust
#[repr(C)]
pub struct PerCpu {
    pub id: usize,
    pub cpu_on_entry: usize,
    pub dtb_ipa: usize,
    pub vcpu_state: VcpuStateCell,
    pub arch_cpu: ArchCpu,
    pub zone: Option<Arc<Zone>>,
    pub ctrl_lock: Mutex<()>,
    pub boot_cpu: bool,
    // percpu stack
}
```


li041:
It seems that those code only support cpu suspending and resuming for riscv. There are just place holders for other archs.

liulog:
> It seems that those code only support cpu suspending and resuming for riscv. There are just place holders for other archs.

You're right, my intention was for each architecture to have its own custom implementation.

You've inspired me; the current implementation isn't strongly architecture-dependent and can be moved to an architecture-independent code location.

li041 (approved):
LGTM.

PR #304, opened 2026-05-09
loongarch64: add SMP support for LoongArch 3A6000

BoneInscri:
## Overview

This PR adds full SMP (Symmetric Multi-Processing) support for LoongArch 3A6000 platforms in hvisor, enabling the hypervisor to boot and manage multiple physical CPUs simultaneously.

### Multi-core Boot (`cpu.rs`, `entry.rs`)

- Added `BootContext` struct to capture the full CPU register state (GPRs + all CSRs including
  TLB, DMW, timer, save registers) at boot time for both primary and secondary CPUs.
- Added `arch_secondary_entry` naked function: secondary CPUs snapshot their full CSR state into
  `CPU_BOOT_CONTEXT_ADDRESS` (a platform-defined shared memory region), then jump into the
  common hvisor entry flow.
- `cpu_start` now wakes secondary CPUs via IPI mailbox, writing the entry address and DTB pointer
  into the per-core mailbox registers.
- Added `GcsrSnapshot` to extract and restore GCSR state from `BootContext` for both primary and
  secondary CPUs.

### IPI Extension for 3A6000 (`ipi.rs`)

- Extended IPI MMIO coverage from 4 cores to 8 cores: added `CORE4_IPI` ~ `CORE7_IPI` mapped to a second MMIO base (`0x8000_0000_1fe1_0000`) for the 3A6000's second CPU cluster.
- Extended all IPI operations (`ipi_write_action`, `mail_send`, `enable_ipi`, `clear_all_ipi`, `get_ipi_status`, `dump_ipi_registers`) to handle cpu_id 4–7.
- Added `HVISOR_SHUTDOWN` IPI action constant.
- Implemented `ipi_send_general()` to map hvisor IPI events (`WAKEUP`, `SHUTDOWN`, `VIRTIO_INJECT_IRQ`, `WAKEUP_VIRTIO_DEVICE`, `CLEAR_IRQ`) to hardware IPI actions, replacing the generic `ipi_write_action()` in `arch_send_event()`.

### EIOINTC Driver (`eiointc.rs`, new file, ~705 lines)

- Full software model of the LoongArch Extended I/O Interrupt Controller (EIOINTC) for virtualization, supporting 256 IRQs routed across up to 256 vCPUs and 8 IP lines.
- Implements IOCSR read/write emulation for all EIOINTC register ranges: nodetype, ipmap, enable, bounce, ISR, coreisr, coremap.
- Bitmap operations (`set_bit`, `clear_bit`, `find_first_bit`) over a 3D array  `[vcpu][ip][word]` for efficient per-CPU interrupt state tracking.
- Handles IRQ injection into guest via `coreisr` updates and IPI dispatch.

### Per-CPU Timer (`timer.rs`, new file)

- Added `restore_timer()` to correctly restore guest timer state on vCPU context switch.
- Handles one-shot vs periodic timer modes, computes remaining `tval` delta based on wall-clock (`ktime_get()`) and the saved expiry timestamp in `arch_cpu.expire`.
- Injects timer interrupt (`INT_TI`) into the guest when a periodic timer fires during restore.

### Trap Handling (`trap.rs`)

- Refactored trap dispatch for SMP: each CPU now independently handles its own trap context.
- Added fault handling improvements: unhandled exceptions now log full context (ecode, esubcode, era, badi, badv) and call `idle()` instead of panicking, aiding debugging on real hardware.

### Boot Method Support (`config.rs`, `zone.rs`, `platform/mod.rs`)

- Added `boot_method` field (`[u8; CONFIG_NAME_MAXLEN]`) to `HvZoneConfig` and `Zone` structs.
- Bumped `CONFIG_MAGIC_VERSION` to `0x6` to stay in sync with hvisor-tool.
- On zone start, if `boot_method` starts with `"acpi"`, registers `a0/a1/a2` are set to pass ACPI boot parameters (`cmd_line_ptr`, `efi_system_table`) to the guest kernel; otherwise they are zeroed for DTB boot.

### Platform Config (`ls3a5000/board.rs`, `ls3a6000/board.rs`)

- Cleaned up `ROOT_ZONE_MEMORY_REGIONS`: removed unused reserved RAM entries, corrected RAM size for the upper region (`0x100000000` to `0x780000000`, covering 30 GB).
- Expanded `ROOT_ZONE_CPUS` from 4 cores (bit 0–3) to 7 cores (bit 0–6) for ls3a6000.
- Updated IPI buffer size: `LoongArch64IpiState.buf` expanded from `[u64; 4]` to `[u64; MAX_CPU_NUM]` to support per-CPU mailbox data for all cores.

### Build Fix (`build.rs`)

- Fixed broken symlink detection when re-linking `__board.rs`: replaced `target_path.exists()` (which returns `false` for broken symlinks) with `target_path.symlink_metadata().is_ok()`, ensuring stale symlinks are always cleaned up before re-creation.

## Tested on

- LoongArch 3A6000 (ls3a6000), booting Linux root zone with 7 physical CPUs


src/arch/loongarch64/cpu.rs
```diff
@@ -13,25 +13,282 @@
...
+            dmw2: b.second_dmw2, dmw3: b.second_dmw3,
+        }
+    }
+
+    fn write_all(&self) {
+        write_gcsr_crmd(self.crmd);
+        write_gcsr_prmd(self.prmd);
+        write_gcsr_pgdh(self.euen); // NOTE: intentional mapping from original code
+        write_gcsr_pgdl(self.misc); // NOTE: intentional mapping from original code
+        write_gcsr_tval(self.ecfg); // NOTE: intentional mapping from original code
```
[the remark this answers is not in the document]
enkerewpo: what's `NOTE: intentional mapping from original code`'s meaning?

enkerewpo:
please run `make fmt` to format your code before PR and update the license headers for all source file (see CI / linter, license-checker

and also please fix the compilation problem shown in CI, thank you👍

src/main.rs
```diff
@@ -82,6 +85,18 @@ static INIT_EARLY_OK: AtomicU32 = AtomicU32::new(0);
 static INIT_LATE_OK: AtomicU32 = AtomicU32::new(0);
 static MASTER_CPU: AtomicI32 = AtomicI32::new(-1);
 
+#[cfg(target_arch = "loongarch64")]
```
enkerewpo: we recommend using less #[cfg(target_arch = "xxx")] in the common logic in hvisor that shared by all platforms

platform/loongarch64/ls3a5000/linker.ld
```diff
@@ -1,5 +1,5 @@
 ENTRY(arch_entry)
-BASE_ADDRESS = 0x90000001f0000000;
+BASE_ADDRESS = 0x90000007f0000000;
```
enkerewpo, line 2: using a 32GB address is not possible on most loongarch machines, maybe we should restrict the hvisor location < 16G?

src/consts.rs
```diff
@@ -28,15 +28,25 @@ use crate::{memory::addr::VirtAddr, platform::BOARD_NCPUS};
 use core::arch::global_asm;
 
 /// Size of the hypervisor heap.
-pub const HV_HEAP_SIZE: usize = 1024 * 1024; // 1 MiB
+#[cfg(target_arch = "loongarch64")]
```
enkerewpo, line 31: we don't add cfg(arch=xxx) in common code like this(`consts.rs`), maybe we should move these to `board.rs`

and why are we increasing these size?

src/hypercall/mod.rs
```diff
@@ -209,7 +209,9 @@ impl<'a> HyperCall<'a> {
             error!("hv_zone_start: cpu {} already on", boot_cpu);
             return hv_result_err!(EBUSY);
         };
+        #[cfg(not(target_arch = "loongarch64"))]
         self.check_cpu_id();
```
enkerewpo, line 220: why skip `check_cpu_id` on loongarch?

platform/loongarch64/ls3a6000/board.rs
```diff
@@ -13,165 +13,249 @@
...
+//     pci_io_base: 0x00008000,
+//     mem32_base: 0x0,
+//     mem32_size: 0x0,
+//     pci_mem32_base: 0x0,
+//     mem64_base: 0x60000000,
+//     mem64_size: 0x20000000,
+//     pci_mem64_base: 0x60000000,
+//     domain: 0x0,
+// }];
+pub const ROOT_PCI_CONFIG: [HvPciConfig; 0] = [];
```
enkerewpo, line 361: why are we removing the hvisor's virtual PCI config and devices? So all PCI devices are visible in this PR's hvisor root linux?

platform/loongarch64/ls3a5000/linker.ld
BoneInscri: 测试机器用的是16GB x 2 内存条

src/consts.rs
BoneInscri: 增加启动上下文后，HEAP大小不够用

platform/loongarch64/ls3a6000/board.rs
BoneInscri: root不需要，现在所有的PCI设备，root linux都是可以看到的，直通的

platform/loongarch64/ls3a5000/platform.mk
```diff
@@ -22,4 +23,9 @@ HVISOR_ENTRY_PA := 0x9000000080000000
 # QEMU_ARGS += -device loader,file="$(zone0_dtb)",addr=0x8f000000,force-raw=on
 
 $(hvisor_bin): elf
-	$(OBJCOPY) $(hvisor_elf) --strip-all -O binary $@
\ No newline at end of file
+	$(OBJCOPY) $(hvisor_elf) --strip-all -O binary $@
+# objdump + hvisor-trap-vector.txt
+	readelf -a $(hvisor_elf) > hvisor-elf.txt
+	loongarch64-linux-gnu-objdump --disassemble $(hvisor_elf) > hvisor.S
```
dallasxy: Previously, the toolchain used for LoongArch was loongarch64-unknown-linux-gnu-. Since loongarch64-linux-gnu- has only been added here, could you please confirm whether a new loongarch64-linux-gnu- toolchain is actually required, or if this is simply a toolchain naming error?

li041:
Thanks for the PR, this is a great step toward SMP support on LoongArch 3A6000.

One suggestion: for some of the `magic constants` and `specific register configuration` , it would be very helpful to add **_references_** (e.g., LoongArch Architecture Manual or 3A6000 hardware documentation).This would make the code easier to review and maintain, and also reduce the risk of subtle mistakes when these values are architecture- or revision-dependent.A short inline comment like “See LoongArch64 Vol. X, section Y” would be very useful.

li041:
I really appreciate the work on this PR.

If possible, could you also mark the already-addressed review comments as resolved and make sure the CI passes before merging?

dallasxy:
the ci for other arch has already passed, but the compile of loongarch still went wrong, please change the loongarch64-linux-gnu-objdump in platform/loongarch64/ls3a5000/platform.mk to loongarch64-unknown-linux-gnu-objdump

agicy:
Closed by #380.

Issue #225, opened 2025-12-15
Unable to run in release mode for aarch64


li041:
## Description

On AArch64, the CPU ID retrieved during early boot is corrupted in non-debug (optimized) builds. The corruption occurs **before SMP bring-up**, indicating that the issue originates in the early boot path rather than later runtime logic.

### Root Cause

At address `0x40400050`, `x17` is used as a loop variable after compiler optimization. In earlier (non-optimized or debug) builds, this register was not used for this purpose.

However:

* `x17` is a **caller-saved** register per the AArch64 ABI
* Its value is assumed to remain intact across calls during early boot
* This assumption is invalid under optimization

As a result, `x17`—which temporarily holds the CPU ID—is overwritten during `boot_pt_init`, leading to corrupted CPU identification.

### Resolution
The issue is resolved by:
* Replacing the use of `x17` with `x19`, which is **callee-saved** according to the AArch64 calling convention

PR #308, opened 2026-05-13
Refine early boot assembly for aarch64

li041:
# Overview
- Fix corrupted boot args `dtbaddr` for `rust_main` in release mode
- Add necessary barrier(`dsb`/`isb`) to early boot assembly.
- btw, add some comments to describe convetions.

## Sync barriers
- Ensure reloading instructions after cache invalidation
- Ensure completion of some instuctions, e.g. `ic iallu`.

## Corrupted `dtbaddr`  

On AArch64, the `dtbaddr` retrieved during early boot may be corrupted in non-debug (optimized) builds.

### Root Cause

- `x18` is a caller-saved register per the AArch64 ABI.
- In hvisor, Its value is assumed to remain intact across calls during early boot.
- This assumption is invalid under optimization which compiler may use those registers to accelerate.
- As a result, x18—which temporarily holds the `dtbaddr`—is overwritten in the release mode.

### Resolution
The issue is resolved by:
 - Replacing the use of `x18` with `x20`, which is callee-saved according to the AArch64 calling convention.

### Note
- This problem is similar to https://github.com/syswonder/hvisor/issues/225.

### DIFF
--- a/src/arch/aarch64/entry.rs
+++ b/src/arch/aarch64/entry.rs
@@ -25,7 +25,8 @@ pub unsafe extern "C" fn arch_entry() -> ! {
     unsafe {
         core::arch::asm!(
             "
-            // x0 = dtbaddr
+            // `arch_entry` receives dtbaddr in `x0`. Keep it in `x20` until `rust_main`.
+            // Note: Only callee-saved registers (x19-x28) are guaranteed to be preserved across function calls.
             mov x20, x0
 
             /* Insert nop instruction to ensure byte at offset 10 in hvisor binary is non-zero.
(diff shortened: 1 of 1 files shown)


agicy:
The replacement of `x18` with `x20` is very appropriate and correctly addresses the register clobbering issue in release mode. 

I suggest taking this opportunity to perform a cleanup of the `arch_entry` assembly block. Upon review, I've noticed that several system register modifications and cache maintenance operations in this block are currently missing necessary architectural barriers (`ISB`/`DSB`). Strictly following the ARM synchronization requirements here would ensure the stability of the boot process across different microarchitectures.

Inquisitor-201 (approved):
LGTM. Verified:

- x18→x20 fix: x20 is callee-saved (x19-x28), preserved correctly through all intermediate calls (, , , , )
- Barrier additions (/) correct per ARM ARM for SCTLR_EL2, icache invalidation, and CSSELR_EL1→CCSIDR_EL1 sequencing
-  uses only caller-saved registers, no preservation issue
- x19/x20 zeroing defensive but harmless
- Comments updated to match actual calling conventions

Good catch on the release-mode register clobber, been bitten by this pattern before (issue #225). The barrier fixes are sound.

Inquisitor-201 (approved):

PR #327, opened 2026-06-28
Add overlap check between MMIO passthrough regions and MMIO intercept handler regions

Solicey:
### Summary

MMIO passthrough regions (MEM_TYPE_IO, mapped in stage-2 page table) and MMIO intercept handler regions (registered via `mmio_region_register()`) have no cross-overlap validation. If both occupy overlapping GPA ranges, the handler silently becomes dead code — the stage-2 mapping prevents any fault, so the handler is never invoked.

### Background

hvisor uses two mechanisms for guest MMIO:

| Mechanism | Registration | Behavior |
|-----------|-------------|----------|
| Passthrough | `pt_init()` → `gpm.insert()` | Direct hardware access, no stage-2 fault |
| Intercept | `mmio_region_register()` → `ZoneInner.mmio` | Stage-2 fault triggers handler |

Each mechanism checks overlaps internally (gpm checks gpm-vs-gpm; mmio checks exact start-address duplicates), but **no cross-check exists between them**. The zone creation flow processes them independently:

```
zone_create()
  pt_init()      → gpm.insert()          (checks gpm-vs-gpm only)
  mmio_init()    → mmio_region_register() (checks exact duplicate only)
  PCI init       → mmio_region_register() (checks exact duplicate only)
```

### Fix

Added bidirectional overlap checks at two insertion points:

**1. `mmio_region_register()` (src/zone.rs)** — When registering a new intercept handler, panic if:
- It overlaps with any existing MMIO handler region (not just exact duplicates).
- It overlaps with any passthrough region already in the stage-2 page table.

**2. `pt_init()` (all four architectures)** — Before inserting a passthrough region into gpm, check whether it overlaps with any already-registered MMIO handler. Overlap triggers a panic.

New utility methods:
- `MMIORegion::is_overlap_with()` in src/memory/mmio.rs — range overlap test
- `MemorySet::is_range_overlap()` in src/memory/mm.rs — public O(log n) query against stage-2 mappings
- `ZoneInner::is_mmio_handler_overlap()` in src/zone.rs — helper for pt_init checks

On loongarch64, the VIRTIO branch in `pt_init()` was reordered (`mmio_region_register` before `gpm.insert`) to avoid a false positive from the VIRTIO trap page.

### Coverage

| Scenario | Caught by |
|----------|-----------|
| Two intercept handler regions overlap | `mmio_region_register()` handler-vs-handler |
| Passthrough inserted before intercept handler | `mmio_region_register()` handler-vs-gpm |
| Intercept handler registered before passthrough | `pt_init()` gpm-vs-handler |
| Same loop: IO entry before VIRTIO entry | `mmio_region_register()` handler-vs-gpm |
| Same loop: VIRTIO entry before IO entry | `pt_init()` gpm-vs-handler |

### Files Changed

- `src/memory/mmio.rs` — add `MMIORegion::is_overlap_with()`
- `src/memory/mm.rs` — add `MemorySet::is_range_overlap()`
- `src/zone.rs` — add checks in `mmio_region_register()` + `is_mmio_handler_overlap()` helper
- `src/arch/x86_64/zone.rs` — add handler overlap check before `gpm.insert()`
- `src/arch/aarch64/zone.rs` — same
- `src/arch/riscv64/zone.rs` — same
- `src/arch/loongarch64/zone.rs` — same + swap VIRTIO order

### Verification

Tested on x86_64 QEMU by adding a VIRTIO region at the same address as the HPET IO passthrough (0xfed0_0000). The fix correctly panics at zone creation:

```
New MMIO handler region MMIORegion { start: 0xfed00000, size: 0x1000 }
overlaps with passthrough region in stage-2 page table
```

All four architectures (x86_64, aarch64, riscv64, loongarch64) compile successfully.

### DIFF
--- a/src/arch/aarch64/zone.rs
+++ b/src/arch/aarch64/zone.rs
@@ -36,6 +36,17 @@ impl Zone {
             }
             match mem_region.mem_type {
                 MEM_TYPE_RAM | MEM_TYPE_IO => {
+                    // Check for overlap with registered MMIO handler regions.
+                    if inner.is_mmio_handler_overlap(
+                        mem_region.virtual_start as GuestPhysAddr,
+                        mem_region.size as _,
+                    ) {
+                        panic!(
+                            "Passthrough region [{:#x}, {:#x}) overlaps with existing MMIO handler",
+                            mem_region.virtual_start,
+                            mem_region.virtual_start as u64 + mem_region.size
+                        );
+                    }
                     inner
                         .gpm_mut()
                         .insert(MemoryRegion::new_with_offset_mapper(
--- a/src/arch/aarch64/zone.rs
+++ b/src/arch/aarch64/zone.rs
@@ -35,34 +35,21 @@ impl Zone {
                 flags |= MemFlags::IO;
             }
             match mem_region.mem_type {
-                MEM_TYPE_RAM | MEM_TYPE_IO => {
-                    // Check for overlap with registered MMIO handler regions.
-                    if inner.is_mmio_handler_overlap(
+                MEM_TYPE_RAM | MEM_TYPE_IO => inner.insert_passthrough_region(
+                    MemoryRegion::new_with_offset_mapper(
                         mem_region.virtual_start as GuestPhysAddr,
+                        mem_region.physical_start as HostPhysAddr,
                         mem_region.size as _,
-                    ) {
-                        panic!(
-                            "Passthrough region [{:#x}, {:#x}) overlaps with existing MMIO handler",
-                            mem_region.virtual_start,
-                            mem_region.virtual_start as u64 + mem_region.size
-                        );
-                    }
-                    inner
-                        .gpm_mut()
-                        .insert(MemoryRegion::new_with_offset_mapper(
-                            mem_region.virtual_start as GuestPhysAddr,
-                            mem_region.physical_start as HostPhysAddr,
-                            mem_region.size as _,
-                            flags,
-                        ))?
-                }
+                        flags,
+                    ),
+                )?,
                 MEM_TYPE_VIRTIO => {
                     inner.mmio_region_register(
                         mem_region.physical_start as _,
                         mem_region.size as _,
                         mmio_virtio_handler,
                         mem_region.physical_start as _,
-                    );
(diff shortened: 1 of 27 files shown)


src/zone.rs
```diff
@@ -223,8 +223,30 @@ impl ZoneInner {
...
+            {
+                panic!(
+                    "New MMIO handler region {:#x?} overlaps with existing handler {:#x?}",
+                    new_region, existing.region
+                );
+            }
+
+            // Check for overlap with passthrough regions in the stage-2 page table.
+            if self.gpm.is_range_overlap(start, size) {
+                panic!(
```
li041, line 242: This turns a recoverable zone-configuration error into a hypervisor hang. `HvZoneStart` already propagates the `HvResult` returned by `zone_create()`, while the panic handler spins forever. A conflicting non-root-zone config submitted by the root zone will therefore wedge the current root vCPU instead of returning `EINVAL`. Please make `mmio_region_register()` return `HvResult` and propagate it, and return `EINVAL` from the `pt_init()` overlap checks rather than panicking.

src/arch/aarch64/zone.rs
```diff
@@ -36,6 +36,17 @@ impl Zone {
             }
             match mem_region.mem_type {
                 MEM_TYPE_RAM | MEM_TYPE_IO => {
+                    // Check for overlap with registered MMIO handler regions.
+                    if inner.is_mmio_handler_overlap(
```
li041, line 40: This only protects mappings created by `pt_init()`. Stage-2 mappings are also added later through `gpm_mut()`; for example, guest PCI BAR writes call `try_insert_quiet()` in `src/pci/pci_handler.rs` without consulting the MMIO handler list. A guest can therefore place a non-MSI-X BAR over an intercepted GPA, create a valid Stage-2 mapping, and make the handler unreachable again. Please route all passthrough mapping insertions through a `ZoneInner` helper that performs this cross-check, or cover every runtime insertion path. This applies to all four architecture copies of this check.

li041 (changes requested):

Solicey:
CI is not available in this pr, so I will open a new one.

PR #358, opened 2026-07-14
fix(loongarch-sriov): add ARI support

ZhongkaiXu:
1. Fix LoongArch accessor implementation and remove outdated DWC-related comments.
2. Add ARI support to ensure the SR-IOV VF BDFs calculated by hvisor match those used by the VM.


src/pci/pci_struct.rs
```diff
@@ -901,17 +935,77 @@ impl VirtualPciConfigSpace {
...
 
-        if total_vfs == 0 || first_vf_offset == 0 || vf_stride == 0 {
+        if total_vfs == 0 {
             return Ok(());
         }
 
+        let ari_forwarding =
+            upstream.is_some_and(|bridge| bridge.enable_ari_forwarding().unwrap_or(false));
+
+        let mut ctrl = self.backend.read(cap_offset + SRIOV_CTRL_OFFSET, 2)? as u16;
```
dallasxy, line 949: Is there a risk of unintentionally clearing other bits here?

src/device/virtio_trampoline.rs
```diff
@@ -71,6 +74,12 @@ pub fn mmio_virtio_handler(mmio: &mut MMIOAccess, base: usize) -> HvResult {
     if need_interrupt == 1 {
         trace!("notify !!!, cpu id is {}", cpu_id);
     }
+    // Guest EOI: drop HWIS before ACK returns, or level-held inject storms vm_interrupt
+    // with INTERRUPT_STATUS==0 → "irq N: nobody cared".
+    #[cfg(target_arch = "loongarch64")]
+    if mmio.is_write && mmio.address == INTERRUPT_ACK {
+        crate::device::irqchip::ls7a2000::clear_hwi_injected_irq_if_needed();
```
dallasxy: Are there any HWIs here that shouldn't be cleared? There are some related changes in hvisor-tool.

src/pci/pci_struct.rs
ZhongkaiXu: Clearing the other bits is intentional here. This runs during hvisor initialization and resets the previous SR-IOV state before rebuilding the control value with the ARI hierarchy and VF Enable bits. We do not rely on any previous VF control state. I’ll add a comment to clarify this.

dallasxy (approved):

Issue #320, opened 2026-06-02
Replace the CPU used by qemu-system-riscv64 from rv64 to rva23s64.


liulog:
Replace the CPU used by qemu-system-riscv64 from `rv64` to `rva23s64`.

Given that more and more hardware will support rva23 in the future, it's becoming a trend. 

For example, the Spacemit-K3. 

Future support for more rva23 features may be necessary, which requires enabling QEMU support.

PR #364, opened 2026-07-21
 feat(riscv): add RVA23 feature for QEMU platforms

liulog:
## Summary:
- Add an rva23 feature and enable it for `qemu-aia` and `qemu-plic`.
- Configure Vector, Svpbmt, CBO, Sstc, and AIA state-enable support.
- Update QEMU CPU models and device trees for RVA23 extensions.
- Preserve existing behavior on non-RVA23 platforms.

## Related:
#320


liulog:
The failure was caused by an outdated QEMU version in the CI; the version needs to be updated to support the `rva23s64` CPU model.

agicy (approved):
LGTM.

PR #365, opened 2026-07-21
fix(loongarch64): repair IRQ delivery and SMP zone startup

li041:
## Summary
- fix LoongArch interrupt/IPI state handling and per-line virtio IRQ deassertion
- harden LoongArch PCI configuration and SR-IOV initialization
- restore the LS3A6000 root/non-root memory layout and current root kernel entry
- support two-vCPU non-root zones with dense guest CPU IDs
- support Python 3.10 in the Kconfig bootstrap environment

## Validation
- built with `make ARCH=loongarch64 BOARD=ls3a6000 MODE=debug all`
- booted the root Linux 6.11.6 zone on physical CPUs 0-1
- started linux1 on physical CPUs 2-3
- logged into linux1 through the virtio console and verified both guest CPUs
- exercised virtio IRQ assertion/deassertion on the LS3A6000 board

### DIFF
--- a/src/arch/loongarch64/cpu.rs
+++ b/src/arch/loongarch64/cpu.rs
@@ -73,15 +73,22 @@ impl ArchCpu {
         for i in 0..32 {
             self.ctx.x[i] = 0;
         }
-        // set all zone's GCSR.CPUID to 0 beacuse linux running on it will believe it's CPU0
-        // - wheatfox 2025.5.20
-        self.ctx.gcsr_cpuid = 0;
+        let zone_cpu_num = this_cpu_data()
+            .zone
+            .as_ref()
+            .map(|zone| zone.read().cpu_num())
+            .unwrap_or(1);
+        self.ctx.gcsr_cpuid = if zone_cpu_num == 1 {
+            0
+        } else {
+            this_cpu_data().id
+        };
         info!(
             "[[CPU virtualization]] CPU{} run@{:#x}",
             self.get_cpuid(),
             self.ctx.sepc
         );
-        info!("loongarch64: @{:#x?}", self);
+        debug!("loongarch64: @{:#x?}", self);
         // step 1: enable guest mode
         // step 2: set guest entry to era
         // step 3: run ertn and enter guest mode
--- a/platform/loongarch64/ls3a5000/board.rs
+++ b/platform/loongarch64/ls3a5000/board.rs
@@ -77,8 +77,11 @@ pub const ROOT_ZONE_MEMORY_REGIONS: &[HvConfigMemoryRegion] = &[
         mem_type: MEM_TYPE_IO,
         physical_start: 0x1fe00000,
         virtual_start: 0x1fe00000,
-        size: 0x1000,
-    }, // uart0
+        size: 0x3000,
+    }, // IOCSR MMIO: uart0, IPI, EXTIOI.
+    // Reference:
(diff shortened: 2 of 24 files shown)


li041:
This draft PR is intended only as a temporary reference and for validation. It is not intended to be merged in its current form. The related changes will be split into multiple focused PRs and submitted separately for review.

DaisyRLight:
I noticed that you seem to be using only 4 cores during SMP debugging. If so, board-level configurations and CPU-related code will need to be modified.
The guest Linux VM encounters issues when restarted after shutdown, which appears to be interrupt-related—such as problems with interrupt cleanup and the timer.

DaisyRLight:
During testing, I triggered the following error. On the surface, it appears to be a page fault occurring on a secondary core; however, this issue has only been observed once so far.

[ERROR 5] (hvisor::panic:24) panic occurred: PanicInfo {
    payload: Any { .. },
    message: Some(
        unhandled instruction: 0b110010000000000010110001100/0x640058c,
    ),
    location: Location {
        file: "src/arch/loongarch64/trap.rs",
        line: 524,
        col: 17,
    },
    can_unwind: true,
    force_no_backtrace: false,
}
[ERROR 7] (hvisor::panic:24) panic occurred: PanicInfo {
    payload: Any { .. },
    message: Some(
        unhandled instruction: 0b110010000000000010110001100/0x640058c,
    ),
    location: Location {
        file: "src/arch/loongarch64/trap.rs",
        line: 524,
        col: 17,
    },
    can_unwind: true,
    force_no_backtrace: false,
}

li041:
> During testing, I triggered the following error. On the surface, it appears to be a page fault occurring on a secondary core; however, this issue has only been observed once so far.

Thanks for the report! I appreciate you bringing this up.
I have not encountered this issue in my testing so far, so it would be very helpful if you could provide a stable reproduction method or any additional details (e.g., workload, boot configuration, or timing conditions) that can help reproduce it.
I will investigate further once there is a reliable way to trigger the issue.

li041:
Thanks for the feedback!

> I noticed that you seem to be using only 4 cores during SMP debugging. If so, board-level configurations and CPU-related code will need to be modified. 

The board-level configuration and CPU-related code need to be adjusted when using a different number of cores. The corresponding SDK and JSON configuration files also need to be updated accordingly. The configuration details can be found in the hvisor [doc](https://hvisor.syswonder.org/chap02/Loongson-3A5000.html).

>The guest Linux VM encounters issues when restarted after shutdown, which appears to be interrupt-related—such as problems with interrupt cleanup and the timer.

For the guest Linux VM restart issue after shutdown, I will investigate it further, especially the interrupt cleanup and timer-related parts. Thanks for pointing this out.

src/arch/loongarch64/cpu.rs
DaisyRLight, line 167: This implementation currently hardcodes IPIs for cores 1–3, which limits SMP support to exactly 4 cores and doesn't scale with `CONFIG_NR_CPUS`.
Additionally, the `_ => panic!()` clause makes the function brittle.

li041: Thanks for catching this issue. The hardcoded IPI mapping problem and the limitation to only 4 cores have already been fixed in https://github.com/syswonder/hvisor/pull/368.

The new implementation supports scalable SMP startup based on the configured CPU count. Thanks again for the helpful review.

li041:
This draft PR is superseded by https://github.com/syswonder/hvisor/pull/374 with the updated implementation, so I’ll close this one to avoid maintaining two parallel changes. Thanks for the discussion and feedback here!

PR #374, opened 2026-08-05
loongarch64: virtualize SMP guest IPI state

li041:
## Summary

  This PR adds virtualized LoongArch IPI support for SMP guests.

  Previously, guest IPI accesses were not modeled as per-vCPU state, and several
  LoongArch paths assumed that a zone always ran on CPU0. This prevented SMP
  guests from reliably starting secondary CPUs and exchanging IPI actions.

  ## Implementation details

  Guest IPI state is maintained per physical CPU using atomic status, enable, and
  mailbox storage. Blocking IPI sends synchronously publish the updated state and
  event notification before returning, without waiting for the guest to clear the
  IPI.

  The virtual IPI line is asserted while pending action bits are nonzero and is
  deasserted only after all pending actions have been cleared.

### DIFF
--- a/src/arch/loongarch64/hypercall.rs
+++ b/src/arch/loongarch64/hypercall.rs
@@ -17,20 +17,12 @@
 use crate::arch::cpu::this_cpu_id;
 use crate::config::HvZoneConfig;
 use crate::config::CONFIG_MAGIC_VERSION;
-use crate::device::virtio_trampoline::MAX_DEVS;
 use crate::hypercall::HyperCall;
 use crate::hypercall::HyperCallResult;
 impl<'a> HyperCall<'a> {
     pub fn hv_ivc_info(&mut self, ivc_info_ipa: u64) -> HyperCallResult {
         warn!("hv_ivc_info is not implemented for LoongArch64");
         HyperCallResult::Ok(0)
-    }
-
-    pub fn wait_for_interrupt(&mut self, irq_list: &mut [u64; MAX_DEVS + 1]) {
-        use crate::device::irqchip::ls7a2000::*;
-        let status = GLOBAL_IRQ_INJECT_STATUS.lock();
-        drop(status);
-        irq_list[0] = 0; // CAUTION: this is a workaround for loongarch64
     }
 
     pub fn hv_zone_config_check(&self, magic_version: *mut u64) -> HyperCallResult {
@@ -60,7 +52,7 @@ impl<'a> HyperCall<'a> {
 
     pub fn check_cpu_id(&self) {
         let cpuid = this_cpu_id();
-        assert_eq!(cpuid, 0);
+        trace!("CPU ID: {} Start Zone", cpuid);
     }
 
     pub fn hv_virtio_get_irq(&self, virtio_irq: *mut u32) -> HyperCallResult {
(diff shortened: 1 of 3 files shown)


li041:
The CI is currently failing because the CPU mappings introduced in PR #374 are missing. This PR depends on the changes in #374, so it should be merged after #374 has been merged.

src/arch/loongarch64/zone.rs
```diff
@@ -587,6 +634,279 @@ fn handle_generic_mmio(mmio: &mut MMIOAccess, base_addr: usize) -> HvResult {
...
+}
+
+pub fn sync_virtual_ipi_line() {
+    let cpu = this_cpu_id();
+    let status = VIRTUAL_IPI_STATUS[cpu].load(Ordering::Acquire);
+
+    if status != 0 {
+        crate::device::irqchip::ls7a2000::inject_irq(12, false);
+    } else {
+        crate::device::irqchip::ls7a2000::clear_injected_irq(12);
```
enkerewpo, line 659: `clear_injected_irq()` is not defined in `ls7a2000`, so the current LoongArch build fails here. Please include the per-line GCSR interrupt clear helper in this PR.

src/arch/loongarch64/zone.rs
```diff
@@ -587,6 +634,279 @@ fn handle_generic_mmio(mmio: &mut MMIOAccess, base_addr: usize) -> HvResult {
...
+    }
+}
+
+fn sync_or_notify_virtual_ipi_line(cpu: usize) {
+    if cpu == this_cpu_id() {
+        sync_virtual_ipi_line();
+    } else {
+        send_event(cpu, SGI_IPI_ID as usize, IPI_EVENT_SEND_IPI);
```
enkerewpo, line 667: This path is not integrated with the current LoongArch event handling. `arch_prepare_send_event()` drains the target event queue, and `arch_check_events()` panics on `IPI_EVENT_SEND_IPI`; the first cross-vCPU guest IPI can therefore discard queued events and then panic. Please include the event queue and handler changes here.

References: [`arch_prepare_send_event()`](https://github.com/syswonder/hvisor/blob/3a109f058827ff949989a9340afc53bb1c33102f/src/arch/loongarch64/ipi.rs#L360-L366), [`arch_check_events()`](https://github.com/syswonder/hvisor/blob/3a109f058827ff949989a9340afc53bb1c33102f/src/arch/loongarch64/ipi.rs#L347-L356).

src/arch/loongarch64/hypercall.rs
```diff
@@ -26,13 +25,6 @@ impl<'a> HyperCall<'a> {
         HyperCallResult::Ok(0)
```
enkerewpo, line 26: Removing this method breaks LoongArch builds because `HyperCall::hv_virtio_inject_irq()` still calls `self.wait_for_interrupt()` unconditionally. Please keep the implementation or gate the call in the same PR.

Reference: [`src/hypercall/mod.rs`](https://github.com/syswonder/hvisor/blob/3a109f058827ff949989a9340afc53bb1c33102f/src/hypercall/mod.rs#L161-L164).

enkerewpo (commented):
Please address these inline issues before merge.

li041:
This draft PR is superseded by https://github.com/syswonder/hvisor/pull/374 with the updated implementation, so I’ll close this one to avoid maintaining two parallel changes. Thanks for the discussion and feedback here!

PR #375, opened 2026-08-07
fix(loongarch64): fix LA release build

weifenjihe:
**Summary**

- fix LoongArch release-mode boot failures caused by early interrupt enabling
- enable global interrupts only after the per-CPU trap context and stack are ready
- explicitly pass the guest context to the return assembly and mark the path as non-returning

**Validation**

- built with `make elf ARCH=loongarch64 BOARD=ls3a6000 MODE=release LOG=info`
- verified the guest context handling in the release disassembly

### DIFF
--- a/src/arch/loongarch64/cpu.rs
+++ b/src/arch/loongarch64/cpu.rs
@@ -114,8 +114,6 @@ impl ArchCpu {
         }
 
         super::trap::_vcpu_return(ctx_addr as usize);
-
-        panic!("loongarch64: ArchCpu::run: unreachable");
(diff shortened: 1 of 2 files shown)


li041 (approved):
Thank you for your contribution. I have verified this change on the LA3A6000 platform. Both root and non-root Linux boot processes have been tested and confirmed to work without any issues. LGTM.

weifenjihe:
@li041  The CI environment issue should be resolved now. Could you please help re-run the failed CI checks? Thanks!

PR #376, opened 2026-08-09
ci(x86_64): add Asterinas zone1 Jenkins test

yydawx:
## Motivation

Add Asterinas zone1 coverage to the existing x86_64 QEMU Jenkins CI while keeping `x86_64/qemu` as the Linux zone1 test flow.

## Approach

Use a dedicated `x86_64/qemu_asterinas` BID for Asterinas instead of mixing Linux and Asterinas in the same BID. The existing `x86_64/qemu` BID remains unchanged and only tests Linux zone1.

`platform_board: qemu` decouples the Jenkins BID from the make/platform board:

- BID `x86_64/qemu_asterinas` is only a test identity and server artifact directory.
- make still uses the real platform `BOARD=qemu` because `platform/x86_64/qemu_asterinas` does not exist in the source tree.
- `artifact_dir: qemu_asterinas` selects the Asterinas artifact directory.

## Artifact layout

`qemu_asterinas` provides:

- `image/kernel/aster-kernel-osdk-bin`
- `image/virtdisk/initramfs.cpio.gz`

The shared `x86_64/qemu` artifacts still provide zone0 files:

- `image/kernel/setup.bin`
- `image/kernel/vmlinux.bin`
- `image/virtdisk/rootfs1.img`

## Changes

- `Jenkinsfile`: add `platform_board` parsing and the `x86_64/qemu_asterinas` matrix cell.
- `jenkins/ci.yaml`: add the dedicated BID and keep Linux zone1 in `x86_64/qemu`.
- `jenkins/ci_config.py`: expose optional `platform_board`.
- `jenkins/ci_runner.py`: map BID to the real platform board, clear `BID` for hvisor make, and add `asterinas_zone1_regression`.
- `jenkins/prepare.sh`: deploy Asterinas kernel/initramfs, Multiboot2 bootloader, and create `zone1_disk.img` with `mkfs.ext2 -b 4096`.
- `jenkins/terminal.py`: add `send_one_by_one()` for sensitive guest consoles.
- `platform/x86_64/qemu/scripts/boot_zone1_asterinas.sh`: start Asterinas zone1 with the Asterinas virtio config.

## QEMU lifecycle

Each BID starts one QEMU instance and runs its own test cases. There is no Linux/Asterinas state sharing and no restart between two guest types inside one BID.

## Verification

- `x86_64/qemu` passes Linux zone1 startup.
- `x86_64/qemu_asterinas` starts Asterinas zone1 and runs `/test/run_regression_test.sh`.
- CI polls Asterinas regression output until it sees both `All regression tests passed` and exit code 0; a non-zero regression exit fails immediately.
- After regression, CI sends `Ctrl-A d` to detach from the Asterinas screen and verifies the zone0 `root@zone0` prompt.
- Jenkins build 8 is fully green, including `x86_64/qemu_asterinas`.

### DIFF
--- a/jenkins/terminal.py
+++ b/jenkins/terminal.py
@@ -392,6 +392,23 @@ class Terminal:
         payload = command.rstrip("\n") + "\n"
         self.backend.write(payload.encode(self.encoding, errors="replace"))
 
+    def send_one_by_one(
+        self,
+        command: str,
+        char_delay: float = 0.02,
+    ) -> None:
+        """Send a command one character at a time for lossy consoles.
+
+        Asterinas' virtio-console currently consumes one byte per interrupt, so
+        pasting a complete command can drop characters. This helper keeps the
+        command readable while avoiding that loss.
+        """
+        self._ensure_open()
+        for char in command.rstrip("\n"):
+            self.backend.write(char.encode(self.encoding, errors="replace"))
+            time.sleep(char_delay)
+        self.backend.write(b"\n")
+
     def read_for(
         self,
         duration: float = 2.0,
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,6 +7,11 @@ def parseBid(String bid) {
     return [arch: parts[0], board: parts[1]]
 }
 
+/** Optional platform_board in ci.yaml maps BID board to make/platform board. */
+def platformBoard(String bid, String configuredBoard = '') {
+    return configuredBoard ?: parseBid(bid).board
+}
+
 def parseCiBuildArgs(cfg) {
     def buildArgs = [:]
     if (!cfg?.build_args) {
@@ -173,7 +178,7 @@ def kconfigSetupShell(String arch, String board) {
             mkdir -p tools/kconfig
             ln -sfn ${env.KCONFIG_VENV} tools/kconfig/.venv
         fi
-        make defconfig ARCH=${arch} BOARD=${board}
+        make defconfig ARCH=${arch} BOARD=${board} BID=
     """
 }
 
@@ -277,6 +282,7 @@ pipeline {
                             'x86_64/ecx-2300f-peg',
                             'x86_64/nuc14mnk',
                             'x86_64/qemu',
+                            'x86_64/qemu_asterinas',
                         )
                     }
                 }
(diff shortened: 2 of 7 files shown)


Solicey:
I think it is OK to add a new BID for x86_64/qemu_asterinas

yydawx:
@Solicey, could you please review the updated PR?

We decouple the Jenkins BID from the make/platform board via `platform_board: qemu`: the BID `x86_64/qemu_asterinas` is only a test identity and artifact directory, while make still uses the real platform `BOARD=qemu`.

If we kept the old rule that the BID board must equal the platform board, and still wanted a dedicated Asterinas BID, we would have to introduce a `platform/x86_64/qemu_asterinas` directory or force `make BOARD=qemu_asterinas`. That would turn a test scenario into a new platform, break the platform abstraction, and add maintenance cost.

Solicey:
It would be better to add more test instructions to show that zone1 works fine. For example, `screen /dev/pts/0` to enter zone1, `ls`, Ctrl A+D to return to zone0. You can refer to qemu/x86_64 for this.

yydawx:
@Solicey, thanks for the review. We updated the CI flow: after entering the Asterinas zone1 console, it now runs `/test/run_regression_test.sh`, waits for `All regression tests passed` and exit code 0, then sends `Ctrl-A d` to detach back to zone0 and verifies the `root@zone0` prompt. Jenkins build 8 is green, including `x86_64/qemu_asterinas`.

yydawx:
@dallasxy, could you please review this PR?

The PR adds a dedicated `x86_64/qemu_asterinas` Jenkins BID while keeping `x86_64/qemu` as the Linux zone1 flow. It decouples the Jenkins BID from the make/platform board via `platform_board: qemu`, and `x86_64/qemu_asterinas` now runs Asterinas zone1 regression tests.

Jenkins build 8 is fully green, including `x86_64/qemu_asterinas`.

jenkins/terminal.py
```diff
@@ -171,6 +171,14 @@ def send(self, command: str) -> None:
         payload = command.rstrip("\n") + "\n"
         self.backend.write(payload.encode(self.encoding, errors="replace"))
 
+    def send_one_by_one(self, command: str, char_delay: float = 0.05) -> None:
+        """Send a command one character at a time for sensitive guest consoles."""
+        self._ensure_open()
+        payload = command + "\n"
+        for char in payload:
+            self.backend.write(char.encode(self.encoding, errors="replace"))
+            time.sleep(char_delay)
```
Solicey: Can you explain the reason for adding a `send_one_by_one` function?

yydawx: Thanks for the question. We found that the current Asterinas virtio-console receives one byte per IRQ, which conflicts with sending the whole command at once via send(). I also observed this in normal use: the Asterinas terminal cannot paste, and screen/tmux also require byte-by-byte input. That is why I added send_one_by_one. After removing it and using the old send(), Jenkins build 10 hung because send() could not deliver the complete command:
http://183.242.33.186:8100/job/hvisor_pr_ci_2/job/PR-376/10/

One alternative is to use the existing send path in a char-by-char loop to build the full command. I currently prefer a single helper because it keeps the terminal abstraction clean.

Which approach should we take: keep send_one_by_one, or use send() in a loop to send one character at a time?

Solicey:
It seems that there are some conflicts with the dev branch, you can resolve these conflicts manually.

Solicey (approved):

PR #377, opened 2026-08-09
feat: add hypercall to query ECAM base

yydawx:
## Motivation

ECAM base can differ across x86_64 devices. hvisor already parses ACPI MCFG, but hvisor-tool needs a way to obtain the correct base before constructing the zone configuration.

## Changes

- `src/hypercall/mod.rs`: add `HvGetEcamBase = 9` hypercall and dispatch it.
- `src/arch/x86_64/hypercall.rs`: implement `hv_get_ecam_base()` to write the ACPI MCFG ECAM base into the caller-provided address.
- aarch64/riscv64/loongarch64 hypercall files: add a no-op handler returning 0 so the shared hypercall dispatch compiles.

## hvisor-tool

hvisor-tool queries this hypercall before starting a zone and overrides `pci_config[0].ecam_base`. See https://github.com/syswonder/hvisor-tool/pull/115.

## Verification

- hvisor x86_64 build passes.
- Tested with zone JSON `ecam_base=0x12345000`; hvisor receives `0xe0000000` after hvisor-tool correction, Asterinas starts and the shell responds to `echo ci-ok`.

### DIFF
--- a/src/arch/aarch64/hypercall.rs
+++ b/src/arch/aarch64/hypercall.rs
@@ -24,6 +24,10 @@ use crate::hypercall::HyperCallResult;
 use crate::zone::this_zone_id;
 
 impl<'a> HyperCall<'a> {
+    pub fn hv_get_ecam_base(&mut self, _ecam_base: *mut u64) -> HyperCallResult {
+        HyperCallResult::Ok(0)
+    }
+
     pub fn hv_ivc_info(&mut self, ivc_info_ipa: u64) -> HyperCallResult {
         let zone_id = this_zone_id();
         let zone = this_zone();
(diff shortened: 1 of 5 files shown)


Solicey:
We'd better not called `apply_dynamic_ecam` inside `hv_zone_start`, because it is a x86-specific function. (We are avoiding things like #[cfg(target_arch = "x86_64")] in shared functions). It might be a good idea to call this function in hvisor-tool. You can add a new hyper-code such as HvGetEcamBase, and a hypercall function which gives hvisor-tool the correct ecam base. Then, hvisor-tool can correct this address and pass it to hvisor.

yydawx:
Thanks for the review. I updated the implementation to follow your suggestion:

- Removed `apply_dynamic_ecam` from the shared `hv_zone_start` path.
- Added an `HvGetEcamBase` hypercall that lets hvisor-tool query the ACPI MCFG ECAM base.
- hvisor-tool now queries this hypercall before starting a zone and overrides `pci_config[0].ecam_base`.

Related hvisor-tool PR: https://github.com/syswonder/hvisor-tool/pull/115.

Verification: with zone JSON `ecam_base=0x12345000`, hvisor receives `0xe0000000` after the hvisor-tool correction, and Asterinas boots to the shell.

Solicey (approved):

PR #380, opened 2026-08-11
fix(loongarch64): make SMP IPI and Virtio IRQ delivery stateful

li041:
## Summary

This PR fixes LoongArch SMP guest startup and interrupt delivery issues by introducing explicit virtualization state for guest CPUs, IPIs, event notifications, and HWI interrupt lines.

The main changes are:

- Add SMP guest CPU initialization based on zone CPU mappings instead of assuming CPU0.
- Virtualize guest IPI status, enable, mailbox, send, and clear operations.
- Improve event notification handling to avoid dropping pending events.
- Track HWI0-HWI7 as independent level-triggered interrupt lines.
- Fix Virtio interrupt injection by deasserting individual IRQ lines instead of using global interrupt state.

## Motivation

The previous LoongArch implementation relied on UP guest assumptions, which caused incorrect behavior for SMP guests:

- `GCSR.CPUID` was always initialized as zero, preventing correct secondary CPU startup.
- Event notification could overwrite or lose pending events during concurrent delivery.
- Physical IPI handling cleared unrelated IPI bits.
- Guest IPI actions were not maintained after injection.
- Virtio IRQ cleanup used global state and could clear unrelated pending interrupts.

These issues resulted in secondary CPU boot failures, lost reschedule/call-function IPIs, spurious interrupts, and Virtio console/block stalls.

### Hypercall ABI

This PR introduces a breaking change to `HvClearInjectIrq`:

- Hypercall number changes from `20` to `11`. The mandatory number change is caused by an ioctl command collision. `HVISOR_SET_BOOT_MODE` already uses ioctl command number `9`:

 ```c
  #define HVISOR_SET_BOOT_MODE \
      _IOW(1, 9, struct hv_zone_boot_mode *)

  #define HVISOR_DEASSERT_IRQ \
      _IOW(1, 9, struct hvisor_irq_line_args)
```
Linux ioctl values encode the direction, type, command number, and argument size. On a 64-bit platform, `struct hv_zone_boot_mode *` is 8 bytes and `struct hvisor_irq_line_args` is also 8 bytes. The two definitions therefore produce the same ioctl value, causing duplicate switch cases in the driver.

### Virtio IRQ Deassert ABI

This PR replaces the previous broadcast interrupt-clear operation with a precise per-IRQ deassert interface.

The previous implementation ignored the IRQ identifier and broadcast a clear event to all running non-root CPUs. This behavior is incorrect for SMP guests because multiple interrupt sources may be pending simultaneously, and clearing global interrupt state can remove unrelated interrupts. The arguments for `HvClearInjectIrq` change to: `arg0: target zone_id`, `arg1: target irq_id`. The operation now deasserts only the specified guest interrupt line.

A matching hvisor-tool update(https://github.com/syswonder/hvisor-tool/pull/111) is required : The driver must pass `(zone_id, irq_id)` to Hvisor.

### Event Notification 

This PR replaces the previous two-stage event notification interface:
```rust
arch_prepare_send_event(cpu_id, ipi_id, event_id);
arch_send_event(cpu_id, ipi_id);
```
with
```
arch_notify_event(cpu_id, ipi_id, event_id, queue_was_empty);
```
The new interface moves event queue state handling into the common event layer. The event is enqueued first, and the architecture layer is informed whether the target queue was previously empty.

This allows different architectures to apply architecture-specific notification strategies:

- AArch64, RISC-V, and x86_64 continue to notify the target CPU for every event.
- LoongArch sends the physical event doorbell only when the queue transitions from empty to non-empty, avoiding redundant IPIs without exposing LoongArch-specific state in the common event layer.

### DIFF
--- a/src/arch/loongarch64/ipi.rs
+++ b/src/arch/loongarch64/ipi.rs
@@ -15,81 +15,61 @@
 //      Yulong Han <wheatfox17@icloud.com>
 //
 use crate::arch::cpu::this_cpu_id;
-use crate::consts::IPI_EVENT_CLEAR_INJECT_IRQ;
-use crate::device::common::MMIODerefWrapper;
+use crate::consts::{IPI_EVENT_CLEAR_INJECT_IRQ, IPI_EVENT_SEND_IPI};
 use core::arch::asm;
 use core::ptr::write_volatile;
-use loongArch64::cpu;
 use loongArch64::register::ecfg::LineBasedInterrupt;
 use loongArch64::register::*;
-use loongArch64::time;
-use tock_registers::fields::FieldValue;
-use tock_registers::interfaces::{ReadWriteable, Readable, Writeable};
-use tock_registers::register_bitfields;
-use tock_registers::register_structs;
-use tock_registers::registers::{ReadOnly, ReadWrite, WriteOnly};
 
 pub fn arch_send_event(cpu_id: u64, sgi_num: u64) {
     debug!(
         "loongarch64: arch_send_event: sending event to cpu: {}, sgi_num: {}",
         cpu_id, sgi_num
     );
-    // just call ipi_write_action
-    ipi_write_action(cpu_id as usize, sgi_num as usize);
+    // Route hvisor event doorbells through the IOCSR sender, whose payload
+    // carries the target CPU ID instead of indexing the fixed legacy table.
+    ipi_write_action_percore(cpu_id as usize, sgi_num as usize);
 }
 
-register_bitfields! [
-  u32,
-  pub IpiStatus [ IPISTATUS OFFSET(0) NUMBITS(32) ],
-  pub IpiEnable [ IPIENABLE OFFSET(0) NUMBITS(32) ],
-  pub IpiSet [ IPISET OFFSET(0) NUMBITS(32) ],
-  pub IpiClear [ IPICLEAR OFFSET(0) NUMBITS(32) ],
-];
-
-register_bitfields! [
-  u64,
-  pub Mailbox0 [ MAILBOX0 OFFSET(0) NUMBITS(64) ],
-  pub Mailbox1 [ MAILBOX1 OFFSET(0) NUMBITS(64) ],
-  pub Mailbox2 [ MAILBOX2 OFFSET(0) NUMBITS(64) ],
-  pub Mailbox3 [ MAILBOX3 OFFSET(0) NUMBITS(64) ],
-];
-
-register_structs! {
-  #[allow(non_snake_case)]
-  pub IpiRegisters {
-    (0x00 => pub ipi_status: ReadOnly<u32, IpiStatus::Register>),
-    (0x04 => pub ipi_enable: ReadWrite<u32, IpiEnable::Register>),
-    (0x08 => pub ipi_set: WriteOnly<u32, IpiSet::Register>),
-    (0x0c => pub ipi_clear: WriteOnly<u32, IpiClear::Register>),
-    (0x10 => _reserved0: [u8; 0x10]),
-    (0x20 => pub mailbox0: ReadWrite<u64, Mailbox0::Register>),
-    (0x28 => pub mailbox1: ReadWrite<u64, Mailbox1::Register>),
-    (0x30 => pub mailbox2: ReadWrite<u64, Mailbox2::Register>),
-    (0x38 => pub mailbox3: ReadWrite<u64, Mailbox3::Register>),
-    (0x40 => @END),
-  }
+pub fn arch_notify_event(cpu_id: u64, sgi_num: u64, event_id: usize, queue_was_empty: bool) {
+    debug!(
+        "loongarch64: notify event: cpu={}, ipi={}, event={}, queue_was_empty={}",
+        cpu_id, sgi_num, event_id, queue_was_empty
+    );
+    if queue_was_empty {
+        arch_send_event(cpu_id, sgi_num);
+    }
 }
 
 const MMIO_BASE: usize = 0x8000_0000_1fe0_0000;
-const IPI_MMIO_BASE: usize = MMIO_BASE;
-const IPI_ANY_SEND_BASE: usize = MMIO_BASE + 0x1158;
+const IOCSR_IPI_STATUS: usize = 0x1000;
+const IOCSR_IPI_ENABLE: usize = 0x1004;
+const IOCSR_IPI_CLEAR: usize = 0x100c;
 
-// IPI registers, use this if you don't want to use the percore-IPI feature
-pub static CORE0_IPI: MMIODerefWrapper<IpiRegisters> =
-    unsafe { MMIODerefWrapper::new(IPI_MMIO_BASE + 0x1000) };
-pub static CORE1_IPI: MMIODerefWrapper<IpiRegisters> =
-    unsafe { MMIODerefWrapper::new(IPI_MMIO_BASE + 0x1100) };
-pub static CORE2_IPI: MMIODerefWrapper<IpiRegisters> =
-    unsafe { MMIODerefWrapper::new(IPI_MMIO_BASE + 0x1200) };
-pub static CORE3_IPI: MMIODerefWrapper<IpiRegisters> =
-    unsafe { MMIODerefWrapper::new(IPI_MMIO_BASE + 0x1300) };
+#[inline]
+fn iocsr_read32(reg: usize) -> u32 {
+    let value: usize;
+    unsafe {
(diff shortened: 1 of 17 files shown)


src/hypercall/mod.rs
```diff
@@ -292,8 +335,11 @@ impl<'a> HyperCall<'a> {
         zone_w.cpu_set().iter().for_each(|cpu_id| {
             let _lock = get_cpu_data(cpu_id).ctrl_lock.lock();
             get_cpu_data(cpu_id).cpu_on_entry = INVALID_ADDRESS;
+            #[cfg(target_arch = "loongarch64")]
+            crate::device::irqchip::ls7a2000::clear_guest_irq_lines(cpu_id);
```
enkerewpo, line 339: For a remote CPU, `clear_guest_irq_lines` enqueues `IPI_EVENT_VIRTIO_INJECT_IRQ` so the target will run `sync_guest_irqs()`. But the `send_event(cpu_id, .., IPI_EVENT_SHUTDOWN)` on the next line goes through `add_event`, which for `IPI_EVENT_SHUTDOWN` does:

```rust
if event_id == IPI_EVENT_SHUTDOWN {
    // shutdown will make cpu idle and won't process any events
    e.clear();
}
```

so the sync event queued a moment earlier is dropped, and that CPU's `GINTC.VIP` keeps whatever bits were asserted. `GUEST_HWI_ASSERTED` is already zeroed by the `swap(0)` inside `clear_guest_irq_lines`, so nothing will re-sync it either.

Is the hardware `GINTC` guaranteed to be re-initialised when the CPU is next started into a zone? If not, the next zone to run on that CPU could see a stale injected HWI.

src/device/irqchip/ls7a2000/mod.rs
```diff
@@ -110,6 +107,94 @@ const INT_PERF: usize = 10;
 const INT_TIMER: usize = 11;
 const INT_IPI: usize = 12;
 
+static GUEST_HWI_ASSERTED: [AtomicU32; MAX_CPU_NUM] = {
+    const C: AtomicU32 = AtomicU32::new(0);
+    [C; MAX_CPU_NUM]
+};
+
+const IRQ_TRACE_LIMIT: usize = 64;
```
enkerewpo: The `[HVDBG:*]` traces here and in `zone.rs` (`trace_virtual_ipi_transition`) are at `info!` level with a hard-coded 64-entry budget. They read like bring-up instrumentation rather than something to keep — could they be dropped, or at least demoted to `debug!` so they compile out of normal builds?

Minor related point: `IRQ_TRACE_SEQ.fetch_add` runs unconditionally on every `sync_guest_irqs()`, i.e. on every virtio interrupt, long after the budget is exhausted.

src/arch/loongarch64/ipi.rs
```diff
@@ -240,57 +270,35 @@ pub fn mail_send(data: usize, cpu_id: usize, mailbox_id: usize) {
...
-        1 => &CORE1_IPI,
-        2 => &CORE2_IPI,
-        3 => &CORE3_IPI,
-        _ => {
-            error!("enable_ipi: invalid cpu_id: {}", cpu_id);
-            return;
-        }
-    };
-    ipi.ipi_enable.write(IpiEnable::IPIENABLE.val(0xffffffff));
+    assert_eq!(cpu_id, this_cpu_id());
```
enkerewpo: `assert_eq!` here (and in `clear_all_ipi`, `clear_ipi_bits`, `get_ipi_status`) panics the hypervisor on a caller mistake.

Since these moved from the per-core MMIO windows to IOCSR, they can only ever act on the current core, so `cpu_id` is now purely a self-check. Either drop the parameter entirely — which makes the constraint impossible to violate — or return an error, rather than taking the whole machine down.

enkerewpo (commented):
Nice cleanup of the LoongArch interrupt path — the SMP CPU-ID mapping, the level-triggered HWI lines, and the per-IRQ deassert are all clear improvements over the previous UP assumptions.

One thing worth calling out because it is easy to get wrong: the doorbell coalescing in `arch_notify_event` looks correct to me. The drain loop in `handle_interrupt` clears the doorbell bit *before* each `fetch`, never after the last one, so a doorbell that lands mid-drain still leaves the bit set and re-fires — the "sender skipped the doorbell because the queue was non-empty" case cannot lose an event. Might be worth a comment there, since the ordering is load-bearing and a later refactor that moves the clear after the fetch would silently break it.

A few smaller things inline. Note also that syswonder/hvisor-tool#111 as it stands still has `HVISOR_HC_CLEAR_INJECT_IRQ 20` and an ioctl number that collides with `HVISOR_SET_BOOT_MODE`, so the two PRs are not yet consistent — I left details on that PR.

src/arch/loongarch64/ipi.rs
```diff
@@ -34,10 +34,19 @@ pub fn arch_send_event(cpu_id: u64, sgi_num: u64) {
         "loongarch64: arch_send_event: sending event to cpu: {}, sgi_num: {}",
         cpu_id, sgi_num
     );
-    // just call ipi_write_action
     ipi_write_action(cpu_id as usize, sgi_num as usize);
 }
 
+pub fn arch_notify_event(cpu_id: u64, sgi_num: u64, event_id: usize, queue_was_empty: bool) {
```
enkerewpo, line 34: Small asymmetry worth noting: `enable_ipi` / `clear_all_ipi` / `clear_ipi_bits` / `get_ipi_status` all moved to IOCSR and are no longer bounded by CPU count, but the notification path underneath `arch_notify_event` still goes through `arch_send_event` → `ipi_write_action`, which dispatches on the fixed `CORE0_IPI`..`CORE3_IPI` table and, for `cpu_id > 3`, only logs:

```rust
_ => {
    error!("ipi_write_action_legacy: invalid cpu_id: {}", cpu_id);
    return;
}
```

Both LoongArch boards are `BOARD_NCPUS = 4` today so this is unreachable, but for a PR whose subject is SMP it would be easy for that to become a silently dropped doorbell later. Would routing the send through `IOCSR_IPI_SEND` (as `ipi_write_action_percore` already does) make sense here?

li041:
> One thing worth calling out because it is easy to get wrong: the doorbell coalescing in `arch_notify_event` looks correct to me. The drain loop in `handle_interrupt` clears the doorbell bit _before_ each `fetch`, never after the last one, so a doorbell that lands mid-drain still leaves the bit set and re-fires — the "sender skipped the doorbell because the queue was non-empty" case cannot lose an event. Might be worth a comment there, since the ordering is load-bearing and a later refactor that moves the clear after the fetch would silently break it.

Added a comment in `handle_interrupt` documenting why the doorbell must be cleared before each queue fetch. The comment explains that a producer racing with the final empty fetch must leave a new doorbell pending so the event is not lost.

src/arch/loongarch64/ipi.rs
li041: Besides, I've removed all those old implementations to avoid wrong usage of those functions which limits cores nums to 4.

src/hypercall/mod.rs
li041: Thansk. I haven't noticed the stale irq states when rebooting.

Fixed this by calling sync_guest_irqs() in the IPI_EVENT_SHUTDOWN handler immediately before entering idle(). This handles the case where IPI_EVENT_VIRTIO_INJECT_IRQ was discarded when the shutdown event cleared the queue, ensuring the cleared software bitmap is still written to GINTC.VIP before the CPU is reused by another zone.

enkerewpo (commented):
All four points look properly addressed, thanks.

- Calling `sync_guest_irqs()` in the `IPI_EVENT_SHUTDOWN` handler is the right place: it runs on the target CPU after `clear_guest_irq_lines` has already zeroed the software bitmap, so `GINTC.VIP` is written back before the CPU can be reused by another zone.
- Dropping the `cpu_id` parameter outright is better than what I suggested — the constraint is now impossible to violate rather than merely checked at run time.
- Routing `arch_send_event` through `ipi_write_action_percore` is equivalent for the doorbell (`ffs(0x8) - 1` yields bit 3, matching `hvisor_mask`), and deleting the legacy four-core table means nothing can silently drop an event on a machine with more cores. I checked that none of the removed functions still have callers.
- The `[HVDBG:*]` traces are gone from both `ls7a2000/mod.rs` and `zone.rs`.

Also confirmed that syswonder/hvisor-tool#111 now uses ioctl command 11 and `HVISOR_HC_CLEAR_INJECT_IRQ 11`, so the two sides agree on the deassert ABI.

LGTM.

enkerewpo (approved):
LGTM

PR #382, opened 2026-08-20
fix(loongarch64): Fix Virtio using too much CPU

weifenjihe:
## Summary

Fix excessive VirtIO backend CPU usage on LoongArch64.

- Enable VirtIO wakeup IPI for LoongArch64.
- Use CPUINTC IRQ 4 (HWI2), matching the device tree.
- Prevent repeated wakeups and HWI update races.
- Allow `hvisor-virtio` to sleep in `epoll_wait()` when idle.

Tested by building hvisor, hvisor-tool, and hvisor.ko for LoongArch64.

### DIFF
--- a/platform/loongarch64/ls3a6000/board.rs
+++ b/platform/loongarch64/ls3a6000/board.rs
@@ -207,7 +207,7 @@ pub const ROOT_ZONE_MEMORY_REGIONS: &[HvConfigMemoryRegion] = &[
     }, // PCI memory resource
 ];
 
-pub const IRQ_WAKEUP_VIRTIO_DEVICE: usize = 32 + 0x20;
+pub const IRQ_WAKEUP_VIRTIO_DEVICE: usize = 4;
 pub const ROOT_ZONE_IRQS_BITMAP: &[BitmapWord] = &get_irqs_bitmap(&[]);
 pub const ROOT_ARCH_ZONE_CONFIG: HvArchZoneConfig = HvArchZoneConfig { dummy: 0 };
 pub const ROOT_ZONE_IVC_CONFIG: [HvIvcConfig; 0] = [];
--- a/src/device/irqchip/ls7a2000/mod.rs
+++ b/src/device/irqchip/ls7a2000/mod.rs
@@ -113,11 +113,15 @@ static GUEST_HWI_ASSERTED: [AtomicU32; MAX_CPU_NUM] = {
     [C; MAX_CPU_NUM]
 };
 
-/// Serializes the software HWI bitmap with GINTC VIP updates for each pCPU.
-static GUEST_HWI_LOCKS: [Mutex<()>; MAX_CPU_NUM] = {
-    const L: Mutex<()> = Mutex::new(());
-    [L; MAX_CPU_NUM]
-};
+/// Serializes the software HWI bitmap with GINTC VIP updates for this pCPU.
+#[percpu::def_percpu]
+static GUEST_HWI_LOCK: Mutex<()> = Mutex::new(());
+
+// The caller ensures the cpu_id is valid.
+#[inline(always)]
+fn get_guest_hwi_lock(cpu: usize) -> &'static Mutex<()> {
+    unsafe { GUEST_HWI_LOCK.remote_ref_raw(cpu) }
+}
 
 fn sync_guest_irqs_unlocked(cpu: usize) {
     use crate::arch::register::gintc;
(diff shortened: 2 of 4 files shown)


li041:
Removing the `VIRTIO_IRQS` lock on LoongArch allows `hv_virtio_inject_irq()` to race with zone shutdown. If a virtio completion is queued after `IPI_EVENT_SHUTDOWN`, the event queue is already non-empty, so no new doorbell is sent. The shutdown handler then enters `idle()` without processing the later event.\n\nWhen the zone is started again, the wakeup event is appended to the non-empty queue and also does not generate a doorbell, leaving the vCPU stopped permanently. Please synchronize virtio completions with shutdown, or explicitly discard and clear pending events before restarting the zone.

platform/loongarch64/ls3a6000/board.rs
```diff
@@ -207,7 +207,7 @@ pub const ROOT_ZONE_MEMORY_REGIONS: &[HvConfigMemoryRegion] = &[
     }, // PCI memory resource
 ];
 
-pub const IRQ_WAKEUP_VIRTIO_DEVICE: usize = 32 + 0x20;
+pub const IRQ_WAKEUP_VIRTIO_DEVICE: usize = 4;
```
li041, line 210: The wakeup IRQ is still `32 + 0x20` on the ls3a5000 and qemu boards. Since `ls7a2000::inject_irq()` rejects IRQs greater than `INT_IPI (12)`, no HWI is injected and the eventfd is never signaled. Guest requests can remain in the shared ring while the daemon sleeps, causing guest MMIO accesses to wait indefinitely.

Please update the wakeup IRQ consistently for all supported LoongArch boards.

src/device/irqchip/ls7a2000/mod.rs
```diff
@@ -112,6 +113,21 @@ static GUEST_HWI_ASSERTED: [AtomicU32; MAX_CPU_NUM] = {
     [C; MAX_CPU_NUM]
 };
 
+/// Serializes the software HWI bitmap with GINTC VIP updates for each pCPU.
+static GUEST_HWI_LOCKS: [Mutex<()>; MAX_CPU_NUM] = {
```
li041: Could we define this as a `#[percpu::def_percpu]` lock and access it with `remote_ref_raw(cpu)` instead? This lock is indexed by the target pCPU, so per-CPU storage matches its ownership and makes the remote-CPU access pattern explicit. The repository already uses the same pattern for `PERCPU_EVENTS` in `src/event.rs` through `get_percpu_events(cpu)`. Keeping this lock consistent with that implementation would also avoid maintaining a separate `[Mutex<()>; MAX_CPU_NUM]` array.

li041 (changes requested):

li041:
I made this follow-up change for two reasons:

1. `GUEST_HWI_ASSERTED` was already accessed under `GUEST_HWI_LOCK` on every path. Using `AtomicU32` therefore did not provide any additional guarantee for the critical section, which must serialize the bitmap update together with the corresponding `GINTC.VIP` update.

2. The name `sync_guest_irqs_unlocked` was misleading. Although the helper itself did not acquire the lock, it could only be called while the caller was already holding the per-CPU lock. This precondition was not documented, making the locking contract easy to misunderstand.

The refactor replaces the atomic bitmap and separate lock with a per-CPU `Mutex<u32>`, and keeps the bitmap update and `GINTC.VIP` synchronization in the same critical section. There is no intended behavior change.

li041:
Thanks for the contribution!

This PR improves LoongArch VirtIO interrupt handling by addressing excessive CPU usage, aligning the LS3A5000 VirtIO IRQ configuration, and moving guest HWI synchronization to per-CPU locking. These changes make interrupt delivery more accurate and reduce unnecessary cross-CPU contention.

Appreciate the work on this!

weifenjihe:
> I made this follow-up change for two reasons:
> 
> 1. `GUEST_HWI_ASSERTED` was already accessed under `GUEST_HWI_LOCK` on every path. Using `AtomicU32` therefore did not provide any additional guarantee for the critical section, which must serialize the bitmap update together with the corresponding `GINTC.VIP` update.
> 2. The name `sync_guest_irqs_unlocked` was misleading. Although the helper itself did not acquire the lock, it could only be called while the caller was already holding the per-CPU lock. This precondition was not documented, making the locking contract easy to misunderstand.
> 
> The refactor replaces the atomic bitmap and separate lock with a per-CPU `Mutex<u32>`, and keeps the bitmap update and `GINTC.VIP` synchronization in the same critical section. There is no intended behavior change.

Thanks for the explanation. I reviewed the refactor, and it matches the original intent. Combining the guest HWI bitmap with the per-CPU mutex makes the locking contract clearer while preserving the existing local synchronization and remote IPI behavior.

li041 (approved):

PR #386, opened 2026-08-24
fix(loongarch64): preserve firmware PCI bridge bus numbers

weifenjihe:
## Issue

On LS3A6000, root Linux stops at:

```text
Waiting for root device /dev/nvme0n1p4...
```

The physical bridge `00:0d.0` uses secondary bus `05`, but `bridge_bus_reg` incorrectly initializes it as `01`, preventing Linux from discovering the NVMe device at `05:00.0`.

## Change

For identity-mapped physical PCI bridges on LoongArch with `CONFIG_NO_PCIE_BAR_REALLOC=y`, initialize `bridge_bus_reg` from the hardware register at offset `0x18`.

Other architectures and remapped PCI bridges are unaffected.

## Verification

- Root Linux successfully discovers the NVMe device and boots

### DIFF
--- a/src/pci/pci_struct.rs
+++ b/src/pci/pci_struct.rs
@@ -352,16 +352,18 @@ pub(crate) fn translate_bridge_bus_reg(
     // Device/function remapping does not change bridge routing. If every bus
     // visible through this bridge keeps its number, preserve the full firmware
     // value, including reserved subordinate ranges and the latency byte.
+    let mut visible_bus_mappings = bus_map
+        .iter()
+        .filter(|config| {
+            config.domain == bdf.domain()
+                && (config.bus == physical_primary
+                    || (config.bus >= physical_secondary && config.bus <= physical_subordinate))
+        })
+        .peekable();
     let identity_bus_map = physical_primary == bdf.bus()
         && bdf.bus() == vbdf.bus()
-        && bus_map
-            .iter()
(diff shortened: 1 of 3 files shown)


src/pci/pci_struct.rs
dallasxy: In get_node(), since set_vbdf() is executed before config_value_init(), the firmware bus number hasn't been cached yet. This causes bridge_bus_reg to be incorrectly set to primary + 1. It would be more reasonable to call config_value_init() right before the final set_vbdf() at the end of get_node()

src/pci/pci_struct.rs
```diff
@@ -321,6 +332,68 @@ impl Debug for Bdf {
...
+    let identity_bus_map = physical_primary == bdf.bus()
+        && bdf.bus() == vbdf.bus()
+        && bus_map
+            .iter()
+            .filter(|config| {
+                config.domain == bdf.domain()
+                    && (config.bus == physical_primary
+                        || (config.bus >= physical_secondary && config.bus <= physical_subordinate))
+            })
+            .all(|config| config.bus == config.v_bus);
```
dallasxy: If the filter result is empty, .all() returns true, which might cause the guest Linux to see an incorrect bus.

dallasxy (approved):

PR #393, opened 2026-09-09
perf(loongarch64): bypass event queue for virtual ipi

### LINKED ISSUE #383: Loongson 3A6000 Real‑Time Stress Test Bug
# Summary
On a Loongson 3A6000 system running root Linux, use the commands hackbench -l -1 & cyclictest -a -m -t -p 99 to perform a stress test for real‑time performance. Within 10 minutes of operation, the cyclictest C parameter (counter) will remain unchanged—meaning that the periodic tasks no longer execute. However, the cyclictest program can still exit normally, and the system remains interactive (not ruling out other hidden bugs).

# Enviornment
Board : Loongarch 3a6000
CPU : core 0 - 1
Cmdline : earlycon console=ttyS0,115200n8 root=/dev/nvme0n1p4 rootwait rw nohlt irqaffinity=0 skew_tick=1

# Test step
Boot the Linux system, log in directly as root without performing any other operations, and then execute the target command
hackbench -l -1 & cyclictest -a -m -t -p 99

li041:
## Summary

  - Update the virtual IPI pending-status bitmap before notifying the target CPU.
  - Use a dedicated physical IPI doorbell (bit 4) for virtual IPI delivery instead of allocating a generic event.
  - Preserve generic event-queue processing for other queued work by draining it only when events are pending.
    - The generic event-queue may be removed by a explicit vcpu abstraction. See https://github.com/syswonder/hvisor/issues/395 for details.

Close #383

  ## Motivation

 Virtual IPI delivery previously allocated a generic event solely to trigger virtual IPI line synchronization. This added unnecessary event-queue traffic and coupled posted virtual IPI state to generic event handling.

This change treats virtual IPI status as posted state: the status bitmap is updated first, and the target CPU is notified through the dedicated doorbell. Generic events remain available for wakeups and other deferred work. However, from a design perspective, the generic event handling mechanism incurs non-negligible overhead, and really requires a refactor.

  ## Verification

  - Hardware validation on the LoongArch 3A6000 board
  - cyclictest with hackbench backend
image

  ## Risks and Limitations
  - The dedicated virtual IPI doorbell and related interrupt handling are currently specific to the LoongArch64 implementation.
  - Generic events still use the per-CPU FIFO and retain its synchronization and processing overhead; this change only removes virtual IPI notifications from that path.


src/arch/loongarch64/trap.rs
```diff
@@ -1280,31 +1287,47 @@ fn handle_interrupt(is: usize) {
...
         }
-        return;
     }
 
     // Handle timer interrupts
-    if is & TIMER_BIT != 0 {
+    if timer_pending {
         debug!("Timer interrupt received");
-        loongArch64::register::ticlr::clear_timer_interrupt();
+    }
```
agicy: Why not combine this if-log statement with the first if-statement in this function?

That is:

```rs
fn handle_interrupt(is: usize) {
    let timer_pending = is & TIMER_BIT != 0;
    if timer_pending {
        // Clear the timer before processing IPI work. A combined IPI+timer
        // exception must not let the IPI path starve the timer indefinitely.
        debug!("Timer interrupt received");
        loongArch64::register::ticlr::clear_timer_interrupt();
    }
    ...
}
```

Since `timer_pending` is used only in the two if-statement mentioned here.

agicy (approved):
LGTM.

